PR #12, opened 2023-11-14
NOSTR calls to relay for signing and browsing

DavidGershony:
Incomplete NOSTR flow but needs to be merged to main before more changes made...


src/Angor/Client/Pages/Create.razor
```diff
@@ -194,6 +260,53 @@
         }
 
         await _RelayService.ConnectToRelaysAsync();
+
```
dangershony, line 263: This will happen every time we land on the create page, is that correct? or we want to connect on landing on the site? is it idempotent?

src/Angor/Client/Pages/Invest.razor
```diff
@@ -352,14 +386,73 @@
...
             {
-                ProjectIdentifier = project.ProjectIdentifier, 
-                InvestmentTransaction = strippedInvestmentTransaction.ToHex(network.Consensus.ConsensusFactory)
-            });
+                ProjectIdentifier = project.ProjectIdentifier,
+                content = encryptedContent,
+                NostrPubKey = project.NostrPubKey,
+                InvestorNostrPrivateKey = nostrPrivateKeyHex
+            }, async _ =>
+                {
```
dangershony, line 408: we should add try catch and show the error message in this block

dangershony: And logs as well

src/Angor/Server/Program.cs
```diff
@@ -32,7 +32,7 @@
 builder.Services.AddSingleton<IInvestmentTransactionBuilder, InvestmentTransactionBuilder>();
 builder.Services.AddSingleton<ISeederScriptTreeBuilder, SeederScriptTreeBuilder>();
 builder.Services.AddSingleton<ITaprootScriptBuilder, TaprootScriptBuilder>();
-
+builder.Services.AddSingleton<ITestNostrSigningFromRelay, TestNostrSigningFromRelay>();
```
dangershony, line 35: add a todo this is a test class

src/Angor/Shared/Services/RelayService.cs
```diff
@@ -181,7 +257,7 @@ private void SetupNostrClient()
 
         private void SetupNostrCommunicator()
         {
-            _nostrCommunicator = new NostrWebsocketCommunicator(new Uri("ws://angor-relay.test"))
+            _nostrCommunicator = new NostrWebsocketCommunicator(new Uri("wss://relay.angor.io"))
```
dangershony, line 260: fetch this from the network service

dangershony (dismissed):

src/Angor/Client/Pages/Create.razor
DavidGershony: It basically checks if there is a connection otherwise opens it.

dangershony (approved):

PR #24, opened 2023-12-06
Handle unconfirmed trx in wallet

dangershony:


src/Angor/Shared/WalletOperations.cs
```diff
@@ -128,10 +128,92 @@ public string GenerateWalletWords()
...
         var signedTransaction = builder.BuildTransaction(true);
-        
+
         return await PublishTransactionAsync(network, signedTransaction);
     }
 
+    private void UpdatePendingLists(AccountInfo accountInfo)
+    {
+        // remove from the pending remove list if it was removed from the indexer
+        var pendingRemove = accountInfo.PendingRemove.ToList();
```
DavidGershony, line 138: Perhaps a name like pending Spent confirmation and pending receive confirmation make it easier to understand what these are

src/Angor/Shared/WalletOperations.cs
```diff
@@ -128,10 +128,92 @@ public string GenerateWalletWords()
...
+        foreach (var addressInfo in accountInfo.AddressesInfo.Concat(accountInfo.ChangeAddressesInfo))
+        {
+            // find all spent inputs to mark them as spent
+            foreach (var utxoData in addressInfo.UtxoData)
+            {
+                foreach (var outPoint in inputs)
+                {
+                    if (utxoData.outpoint.ToString() == outPoint.ToString())
+                    {
+                        if (accountInfo.PendingRemove.All(_ => _.outpoint != utxoData.outpoint))
```
DavidGershony, line 186: Don't you need to call to string on the outpoints?

src/Angor/Shared/Models/AccountInfo.cs
```diff
@@ -11,6 +11,9 @@ public class AccountInfo
     public List<AddressInfo> AddressesInfo { get; set; } = new();
     public List<AddressInfo> ChangeAddressesInfo { get; set; } = new();
 
+    public List<UtxoData> PendingAdd { get; set; } = new();
+    public List<UtxoData> PendingRemove { get; set; } = new();
```
DavidGershony, line 15: We duplicate the data in storage, can't we just set it as a property on the UTXO data that is held in the address info utxo data?

src/Angor/Client/Pages/Wallet.razor
```diff
@@ -594,7 +594,16 @@
 
             var wallet = _walletStorage.GetWallet();
 
-            return await _walletOperations.SendAmountToAddress(wallet, _sendInfo);
+            var res = await _walletOperations.SendAmountToAddress(wallet, _sendInfo);
+
+            if (res.Success)
+            {
+                _walletOperations.UpdateAccountInfoWithSpentTransaction(localAccountInfo, res.Data, _sendInfo.ChangeAddress);
+                localAccountInfo.CalculateBalance();
```
DavidGershony, line 602: This does not return anything

src/Angor/Shared/WalletOperations.cs
dangershony: good catch

src/Angor/Shared/Models/AccountInfo.cs
dangershony: I tried that but it go more complicated, the duplication is negligible anyway

src/Angor/Client/Pages/Wallet.razor
dangershony: It sets the TotalBalance and TotalUnConfirmedBalance properties that are used in the page

dangershony:
this is now ready to review @DavidGershony

src/Angor/Shared/WalletOperations.cs
```diff
@@ -388,9 +476,19 @@ public async Task<(string address, List<UtxoData> data)> FetchUtxoForAddressAsyn
...
+            {
+                _logger.LogWarning($"utxo scan for address {address} was stopped after the limit of {maxutxo} was reached.");
+                break;
+            }
+                
+
             offset += limit;
         } while (true);
 
+        // todo: dan - this is a hack until the endpoint offset is fixed
```
DavidGershony, line 489: Workaround

src/Angor/Shared/WalletOperations.cs
```diff
@@ -156,6 +237,11 @@ public List<UtxoDataWithPath> FindOutputsForTransaction(long sendAmountat, Accou
                      .OrderBy(o => o.utxo.blockIndex)
                      .ThenByDescending(o => o.utxo.value))
         {
+            if (unconfirmedInfo.AccountPendingSpent.Any(p => p.outpoint.ToString() == utxoData.utxo.outpoint.ToString()))
+            {
+                continue;
+            }
```
DavidGershony, line 243: Why not add it as a Where in the Utxos..SelectMany

src/Angor/Shared/WalletOperations.cs
```diff
@@ -128,10 +127,92 @@ public string GenerateWalletWords()
...
+
+        var outputs = transaction.Outputs.AsIndexedOutputs();
+        var inputs = transaction.Inputs.Select(_ => _.PrevOut).ToList();
+        
+        foreach (var addressInfo in accountInfo.AddressesInfo.Concat(accountInfo.ChangeAddressesInfo))
+        {
+            // find all spent inputs to mark them as spent
+            foreach (var utxoData in addressInfo.UtxoData)
+            {
+                foreach (var outPoint in inputs)
```
DavidGershony, line 181: 3 nested loops, isn't there a better way to get the utxos?

src/Angor/Shared/Models/Outpoint.cs
```diff
@@ -17,4 +17,14 @@ public OutPoint ToOutPoint()
     {
         return new OutPoint(uint256.Parse(transactionId), outputIndex);
     }
+
+    public static Outpoint FromOutPoint(OutPoint outPoint)
+    {
+        return new Outpoint { outputIndex = (int)outPoint.N, transactionId = outPoint.Hash.ToString() };
+    }
+
+    public static Outpoint Create(string trxid, int index)
```
DavidGershony, line 26: Shouldn't this just be a constructor?

src/Angor/Client/Pages/Wallet.razor
```diff
@@ -305,8 +312,8 @@
                         </tr>
                     </thead>
                     <tbody>
-                    
-                        @foreach (var addressInfo in localAccountInfo.AddressesInfo.Union(localAccountInfo.ChangeAddressesInfo))
+
+                        @foreach (var addressInfo in accountBalanceInfo.AccountInfo.AddressesInfo.Union(accountBalanceInfo.AccountInfo.ChangeAddressesInfo))
```
DavidGershony, line 316: Would be easier if it was a method on the account balance info

src/Angor/Shared/Models/AccountBalanceInfo.cs
```diff
@@ -0,0 +1,27 @@
...
+public class AccountBalanceInfo
+{
+    public long TotalBalance { get; set; }
+    public long TotalUnconfirmedBalance { get; set; }
+
+
+    public AccountInfo AccountInfo { get; private set; } = new ();
+    public UnconfirmedInfo UnconfirmedInfo { get; private set; } = new ();
+
+    public static AccountBalanceInfo GetBalance(AccountInfo account, UnconfirmedInfo unconfirmedInfo)
```
DavidGershony, line 12: Why is it a static method?

src/Angor/Client/Pages/Wallet.razor
```diff
@@ -596,8 +607,20 @@
...
+            var accountInfo = storage.GetAccountInfo(network.Name);
+            var unconfirmedInfo = _cacheStorage.GetUnconfirmedInfo();
+
+            var res = await _walletOperations.SendAmountToAddress(wallet, _sendInfo);
 
-            return await _walletOperations.SendAmountToAddress(wallet, _sendInfo);
+            if (res.Success)
+            {
+                _walletOperations.UpdateAccountUnconfirmedInfoWithSpentTransaction(accountInfo, unconfirmedInfo, res.Data);
+                accountBalanceInfo = AccountBalanceInfo.GetBalance(accountInfo, unconfirmedInfo);
```
DavidGershony, line 618: The name suggests that the account balance info holds the state end returns the balance not that it is a static method

src/Angor/Client/Pages/Recover.razor
```diff
@@ -662,6 +721,10 @@
             if (!response.Success)
                 return response;
 
+            var unconfirmedInfo = _cacheStorage.GetUnconfirmedInfo();
+            unconfirmedInfo.AddInputsAsPending(endOfProjectTransaction);
+            _cacheStorage.SetUnconfirmedInfo(unconfirmedInfo);
```
DavidGershony, line 726: If you have the same logic in a few methods in this file better have it as a private method so it's easier to understand what happens here.

src/Angor/Shared/WalletOperations.cs
dangershony: ah yes I should fix this when we deploy the fix to the indexer but it needs to be tested on the indexer first

src/Angor/Shared/WalletOperations.cs
dangershony: good idea done

src/Angor/Shared/WalletOperations.cs
dangershony: Even if I use linq there will still be loops it will just look shorter.
To optimize here we will need to use dictionaries but I think this is not really needed

src/Angor/Client/Pages/Wallet.razor
dangershony: good point I will add that

dangershony: this should not even be Union it should be a concat I think

src/Angor/Shared/Models/Outpoint.cs
dangershony: I think its simpler to understand as it is used in linq queries

src/Angor/Client/Pages/Recover.razor
dangershony: Good point

src/Angor/Shared/Models/AccountInfo.cs
DavidGershony: Because of duplicates you need to pass both objects to the wallet operations and check each output you try to use to make sure it's not already used, this seems like you try to avoid a problem only to get it again down the line.

dangershony: The issue is more complicated, if we start to change the state that we receive form the indexer then I will have to look at the state when syncing, the proposed flag on the object will get overridden when we refresh but the utxo is not confirmed yet, though we could overcome that by changing the structure of how we store data.

DavidGershony (dismissed):

DavidGershony (dismissed):

PR #25, opened 2023-12-13
Connect Angor to UniSat Wallet

miladsoft:


src/Angor/UniSatWallet/Exceptions/NoUniSatWalletException.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace UniSatWallet.Exceptions
```
DavidGershony, line 7: If we have a project for each wallet that could become a very big solution, perhaps a project for wallet connectors with a connector and models file per wallet?

DavidGershony (commented):
We need to discuss the project per wallet thing..

dangershony:
I wonder if we should hold this PR till we get the platform to a stable point, though it is all nicely isolated to its own project

dangershony:
Closing for now as this is lower priority

PR #26, opened 2023-12-13
Change NOSTR client to multi connections

DavidGershony:


src/Angor/Client/Pages/Invest.razor
```diff
@@ -472,6 +472,9 @@
 
     private async Task HandleSignatureReceivedAsync(string? nostrPrivateKeyHex, string _)
     {
+        if (recoverySigs?.Signatures.Any() ?? false) //multiple relays for the same message
```
dangershony, line 475: I wonder if we should compare sigs (probably no need) we filter using pub key which is unique per user per project right?

src/Angor/Client/Pages/Settings.razor
```diff
@@ -96,7 +97,8 @@
                 {
                     <tr>
                         <td><a href="@relay.Url" target="_blank">@relay.Url</a></td>
-                            <td style="color: @(relay.Status == UrlStatus.Online ? "green" : relay.Status == UrlStatus.NotReady ? "yellow" :  "red");">@relay.Status.ToString()</td>
+                        <td>@relay.Name</td>
```
dangershony, line 100: A relay has a name? is it assigned or fetched form the relay?

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -172,43 +148,57 @@
...
                     pendingSignature.AmountToInvest = investorTrx.Outputs.AsIndexedOutputs().Skip(2).Take(investorTrx.Outputs.Count - 3) //Todo get the actual outputs with taproot type
                         .Sum(_ => _.TxOut.Value);
-                    
-                    StateHasChanged();
                 }
                 catch (Exception e)
                 {
                     Console.WriteLine(e);
-                    Console.WriteLine(pendingSignature.TransactionHex);
+                    _Logger.LogDebug(pendingSignature.TransactionHex);
```
dangershony, line 155: probably not logging the correct thing here?

src/Angor/Shared/DerivationOperations.cs
```diff
@@ -216,6 +216,22 @@ public Key DeriveProjectNostrPrivateKey(WalletWords walletWords, int index)
 
         return extKey.PrivateKey;
     }
+    
+    public async Task<Key> DeriveProjectNostrPrivateKeyAsync(WalletWords walletWords, int index)
+    {
+        // founder key is derived from the path m/5'
+        ExtKey extendedKey = GetExtendedKey(walletWords);
+
+        var path = $"m/44'/1237'/{index}/0/0";
```
dangershony, line 225: why 1237?

src/Angor/Shared/Models/NostrRelayInfo.cs
```diff
@@ -0,0 +1,21 @@
+using System.Text.Json.Serialization;
+
+namespace Angor.Shared.Models;
+
+public class NostrRelayInfo
+{
+    [JsonPropertyName("name")]
+    public string Name { get; set; }
```
dangershony, line 8: ah ok yes we fetch this data from the relay?

dangershony (approved):
Looks ok to me

src/Angor/Client/Pages/Invest.razor
DavidGershony: The sigs will be different for the same user and project as well

src/Angor/Client/Pages/Settings.razor
DavidGershony: Yes, I have updated the code to do that

src/Angor/Client/Pages/Signatures.razor
DavidGershony: It's only for debugging, I will remove it..

src/Angor/Shared/DerivationOperations.cs
DavidGershony: NOSTR protocol

PR #27, opened 2023-12-14
Search for approved invesetment requests and update state

DavidGershony:


src/Angor/Client/Pages/Signatures.razor
```diff
@@ -172,43 +157,99 @@
...
                     pendingSignature.AmountToInvest = investorTrx.Outputs.AsIndexedOutputs().Skip(2).Take(investorTrx.Outputs.Count - 3) //Todo get the actual outputs with taproot type
                         .Sum(_ => _.TxOut.Value);
-                    
-                    StateHasChanged();
                 }
                 catch (Exception e)
                 {
                     Console.WriteLine(e);
-                    Console.WriteLine(pendingSignature.TransactionHex);
+                    _Logger.LogDebug(pendingSignature.TransactionHex);
```
dangershony, line 164: There should be more text around logs, later we wont be able to know where this came from.

src/Angor/Shared/DerivationOperations.cs
```diff
@@ -216,6 +216,22 @@ public Key DeriveProjectNostrPrivateKey(WalletWords walletWords, int index)
...
     }
+    
+    public async Task<Key> DeriveProjectNostrPrivateKeyAsync(WalletWords walletWords, int index)
+    {
+        // founder key is derived from the path m/5'
+        ExtKey extendedKey = GetExtendedKey(walletWords);
+
+        var path = $"m/44'/1237'/{index}/0/0";
+
+        var task = Task.Run(() => extendedKey.Derive(new KeyPath(path)));
```
dangershony, line 227: this doesn't help in anyway to make it faster, drop the Task here if you ask me it is just confusing.

dangershony (approved):

src/Angor/Shared/DerivationOperations.cs
DavidGershony: Not faster but the UI doesn't freeze when it runs in a task

dangershony: But later down you await the task is it means the browser will still freeze no?

dangershony (approved):

PR #38, opened 2024-01-25
Feat font

sondreb:
Changed to a cleaner, but still unique heading font:

image


dangershony:
The whole UI will change

PR #42, opened 2024-02-02
When recovering coins only set the trx if it was accepted to the chain

dangershony:


src/Angor/Client/Pages/View.razor
```diff
@@ -88,7 +88,7 @@
                         int currentDays = (DateTime.Now - project.ProjectInfo.StartDate).Days;
                         currentDays = currentDays < 0 ? 0 : currentDays;
                         currentDays = currentDays > totalDays ? totalDays : currentDays;
-
+                        totalDays = totalDays > 0 ? totalDays : 1;
```
DavidGershony, line 91: What is this supposed to do?

PR #43, opened 2024-02-02
Fix create bug when publish failed

dangershony:


src/Angor/Client/Pages/Create.razor
```diff
@@ -427,6 +425,8 @@
             return;
         }
 
+        project.ProjectInfo.CreationTransactionId = signedTransaction.GetHash().ToString();
```
DavidGershony, line 428: You also need to handle the new state in line 266 when checking on loading the page

dangershony: This is the result from the relay, what state do we need to handle here?

DavidGershony: We should only navigate to the view page if creation transaction id is populated

dangershony: Gotcha ill fix it

DavidGershony (approved):

src/Angor/Client/Pages/Create.razor
```diff
@@ -252,18 +253,71 @@
...
+
+                        project.ProjectInfo = nostrProject;
+                    }
+
+                    if (applicationDataOnNostr && nostrProfileCreated)
+                    {
+                        var findProject = storage.GetFounderProjects().FirstOrDefault(p => p.ProjectInfo.ProjectIdentifier == project.ProjectInfo.ProjectIdentifier);
+                        if (findProject != null)
+                        {
+                            if (string.IsNullOrEmpty(findProject.CreationTransactionId))
```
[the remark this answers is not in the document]
dangershony: As agreed lets do it in 3 stages, this will come in a following PR

PR #44, opened 2024-02-02
Unifying the wallet database

dangershony:


src/Angor/Client/Pages/Invest.razor
```diff
@@ -383,16 +383,16 @@
             feeData.FeeEstimations.Fees.AddRange(fetchFees);
             feeData.SelectedFeeEstimation = feeData.FeeEstimations.Fees.First();
 
-            var investorKey = _derivationOperations.DeriveInvestorKey(_walletStorage.GetWallet(), project.ProjectInfo.FounderKey);
+            var investorKey = _derivationOperations.DeriveInvestorKey(_walletStorage.GetWallet().WalletWords, project.ProjectInfo.FounderKey);
```
DavidGershony, line 386: We call get wallet 3 times in the same method, can you change it to one call?

src/Angor/Client/Pages/Recover.razor
```diff
@@ -461,7 +461,7 @@
             
             var accountInfo = storage.GetAccountInfo(network.Name);
 
-            var investorPrivateKey = _derivationOperations.DeriveInvestorPrivateKey(_walletStorage.GetWallet(), project.ProjectInfo.FounderKey);
+            var investorPrivateKey = _derivationOperations.DeriveInvestorPrivateKey(_walletStorage.GetWallet().WalletWords, project.ProjectInfo.FounderKey);
```
DavidGershony, line 464: We call get wallet 2 times in the same method here..

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -278,13 +278,13 @@
     {
         var operationResult = await notificationComponent.LongOperation(async () =>
         {
-            var key = DerivationOperations.DeriveFounderRecoveryPrivateKey(_walletStorage.GetWallet(), FounderProject.ProjectIndex);
+            var key = DerivationOperations.DeriveFounderRecoveryPrivateKey(_walletStorage.GetWallet().WalletWords, FounderProject.ProjectIndex);
```
DavidGershony, line 281: We also call the storage 2 times in this method

src/Angor/Shared/Models/Wallet.cs
```diff
@@ -1,6 +1,12 @@
 using System.Text.Json;
 
 namespace Angor.Shared.Models;
+public class Wallet
```
DavidGershony, line 4: I don't think we should have many classes in one file

src/Angor/Client/Pages/Invest.razor
dangershony: Good point in fact this can be a point if improving perf, we parse the wallet every time we pass the words, instead we could derive the parent key once and pass that down the method

DavidGershony: You mean keep it in the keys in the wallet?

dangershony: yes, or derive it once and pass that to the method, right now we derive it each time in all 3 cases.

DavidGershony (approved):

PR #49, opened 2024-03-11
Add password encryption

dangershony:


src/Angor/Client/wwwroot/lib/js-base64/base64.mjs
```diff
@@ -0,0 +1,294 @@
+/**
```
dangershony, line 1: @sondreb any chance you can review this part of the code? especially around password generator and the way I imported the base64 js files

src/Angor/Client/Pages/CheckTransactionCode.razor
```diff
@@ -130,7 +130,7 @@ else
         localAccountInfo = GetAccountInfoFromStorage();
 
         ExtPubKey accountExtPubKey = ExtPubKey.Parse(localAccountInfo.ExtPubKey, network);
-        var walletWords = _walletStorage.GetWallet().WalletWords;
+        var walletWords = new WalletWords();// this is test code we can probably delete it _walletStorage.GetWallet().WalletWords;
```
DavidGershony, line 133: What do you mean by test code we can probably delete?

dangershony: Check transaction code page is a test page no?
We can delete it because we made it to check we can sign a trx in wasm

dangershony:
FYI for the password generator components we should also consider this issue
https://github.com/block-core/blockcore-wallet/issues/232

src/Angor/Client/Pages/Wallet.razor
```diff
@@ -713,27 +815,40 @@
...
+        StateHasChanged();
     }
 
     private async Task SendCoins()
     {
+        if (!passwordComponent.HasPassword())
+        {
+            sendConfirmSpinner = false;
+            sendConfirmModal = false;
+            notificationComponent.ShowErrorMessage("no wallet password");
```
DavidGershony, line 843: The error message is not very user friendly

src/Angor/Client/Pages/Wallet.razor
```diff
@@ -775,7 +891,7 @@
         }
     }
 
-    private void FeeRangeChanged(ChangeEventArgs e)
+    private async Task FeeRangeChanged(ChangeEventArgs e)
```
DavidGershony, line 901: Why did the method change if we don't await anything?

DavidGershony (approved):

PR #51, opened 2024-03-13
General style change

miladsoft:
Setting the style of the controls and the layout of the controls has not been done yet, and in order to work in parallel, I have sent this PR. I continue to work in a new branch.


dangershony:
This PR mixes too many changes, as agreed with @miladsoft we must break it down to smaller PRs

PR #61, opened 2024-03-28
Adding support for mainnet

dangershony:
**Before this PR is pushed we should create a mainnet Angor key**

The logic in going to mainnet or testnet is dependent on the url of the domain.
If it contains "test" or "localhost" the app will default to testnet otherwise mainnet.
However it is possible to force a network in the settings page.

### DIFF
--- a/src/Angor/Client/NetworkConfiguration.cs
+++ b/src/Angor/Client/NetworkConfiguration.cs
@@ -8,29 +8,75 @@ namespace Angor.Client;
 public class NetworkConfiguration : INetworkConfiguration
 {
     public static string AngorTestKey = "tpubD8JfN1evVWPoJmLgVg6Usq2HEW9tLqm6CyECAADnH5tyQosrL6NuhpL9X1cQCbSmndVrgLSGGdbRqLfUbE6cRqUbrHtDJgSyQEY2Uu7WwTL";
+    public static string AngorMainKey = "todo: add an Angor key here";
+
+    private Network currentNetwork;
+
+    public void SetNetwork(Network network)
+    {
+        currentNetwork = network;
+    }
 
     public Network GetNetwork()
     {
-        return new BitcoinSignet();
+        if (currentNetwork == null)
+        {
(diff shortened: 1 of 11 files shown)


src/Angor/Client/NetworkConfiguration.cs
```diff
@@ -8,29 +8,75 @@ namespace Angor.Client;
...
+        }
 
-        return new SettingsUrl { Name = "", Url = "https://explorer.angor.io/tbtc/explorer" };
+        throw new ApplicationException("Network not set");
     }
 
     public List<SettingsUrl> GetDefaultIndexerUrls()
     {
-        return new List<SettingsUrl>
+        if (currentNetwork.NetworkType == NetworkType.Testnet)
```
DavidGershony, line 62: Why not just replace the subdomain in the URL if the rest stays the same?

src/Angor/Client/Pages/Settings.razor
```diff
@@ -58,6 +59,17 @@
...
                                     <p class="text-danger-emphasis">Please tick the box to wipe all storage!</p>
                                 }
+                                
+                                <hr />
+
+                                <!-- Dropdown for network selection -->
+                                <div class="mb-3">
+                                    <label for="networkSelection">Select Network:</label>
+                                    <select class="form-control" id="networkSelection" @bind="selectedNetwork">
+                                        <option value="testnet" selected>Testnet</option>
```
DavidGershony, line 69: Could this be data driven from the network configuration class?

DavidGershony (approved):

src/Angor/Client/NetworkConfiguration.cs
dangershony: it may not always be the case, it is just simpler this way.

src/Angor/Client/Pages/Settings.razor
dangershony: It can but I think it is not urgent rn, we will for now just have main/test.

PR #93, opened 2024-05-02
Change wallet extra passphrase text

dangershony:


sondreb (approved):
This is good, should be additional warning that user must BACKUP this extra word and that setting this word, means that the secret recovery phrase is NOT enough to recover the wallet in the future.

dangershony:
The warning is now 

> You've backed up your wallet words and passphrase securely, as losing them means losing access to your wallet.

PR #95, opened 2024-05-05
Add cy tests

itailiors:


itailiors:
Summary of the PR:

Upgrade cypress version,
Add tests for recover wallet, and for send funds feature 
(Were using a existing account on testnet you can see in enums, because of 3 reasons
1. The time it would take to get test coins would slow down the tests alot.
2. To check that recover wallet works
3. Getting test coins can be flaky)

Added some commands for better readability, and reformatted the code

dangershony (approved):

dangershony:
so the test wallet is gonna run out of coins at some point?

PR #97, opened 2024-05-22
Unit tests for walletOperations, and cypress test

itailiors:


itailiors:
Unit Tests for WalletOperations:

Added tests for wallet word generation, transaction handling with sufficient and insufficient funds, unspent output retrieval, and transaction fee calculation.

Cypress Tests Enhancements:

improved the browseProjectsSpec suite with wallet setup and investment attempt scenarios.

src/Angor.Test/WalletOperationsTest.cs
```diff
@@ -183,10 +202,249 @@ public void AddInputsAndSignTransaction()
...
+        // Assert
+        Assert.Single(coins);
+        Assert.Single(keys);
+        Assert.Equal((uint)0, coins[0].Outpoint.N);
+        Assert.Equal(1500000, coins[0].Amount.Satoshi);
+        Assert.Equal(expectedExtKey.Derive(new KeyPath("m/0/0")).PrivateKey, keys[0]);
+    }
+
+    [Fact]
+    public void CalculateTransactionFee_WithMultipleScenarios()
```
itailiors, line 370: @dangershony glad if you could go over the logic of this test,
mostly on the values, and if there is no need for Assert.Equal(0.00000001m, calculatedFeeSufficient);

src/Angor.Test/WalletOperationsTest.cs
```diff
@@ -183,10 +202,249 @@ public void AddInputsAndSignTransaction()
...
+                    HdPath = "m/0/0"
+                }
+            }
+        }
+        };
+
+        // Act & Assert for sufficient funds
+        var calculatedFeeSufficient = walletOperations.CalculateTransactionFee(sendInfoSufficientFunds, accountInfo, feeRate);
+        Assert.True(calculatedFeeSufficient > 0);
+        Assert.Equal(0.00000001m, calculatedFeeSufficient);  // Assuming an expected fee for validation
```
dangershony, line 417: this is fine, the fee is exactly the same every time because the size of the trx is fixed and feerate is fixed, so it is fine to keep it.

dangershony (approved):

PR #100, opened 2024-06-02
Improve spend funds data

itailiors:


src/Angor/Client/Pages/Signatures.razor
```diff
@@ -376,6 +380,60 @@
         }
     }
 
+    private async Task ApproveAllSignatures()
```
itailiors, line 383: @dangershony is there a better way to do this?
i added another function because approvesignature() resets the password each time, and i didn't want to add a boolean in that function.

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -376,6 +380,60 @@
...
+        var pendingSignatures = signaturesRequests.Where(s => s.TransactionHex != null && s.AmountToInvest != null && s.TimeApproved == null).ToList();
+        int totalSignatures = pendingSignatures.Count;
+
+        for (int i = 0; i < totalSignatures; i++)
+        {
+            var signature = pendingSignatures[i];
+
+            var operationResult = await notificationComponent.LongOperation(async () =>
+            {
+                var words = await passwordComponent.GetWalletAsync();
```
dangershony, line 394: Lok at other pages to see how we have a double method for password request and then we call in to the method that has the spinner

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -376,6 +380,60 @@
...
+
+                var signatureInfo = signProject(signature.TransactionHex, FounderProject.ProjectInfo, Encoders.Hex.EncodeData(key.ToBytes()));
+
+                var sigJson = JsonSerializer.Serialize(signatureInfo, RelayService.settings);
+
+                var nostrPrivateKey = await DerivationOperations.DeriveProjectNostrPrivateKeyAsync(words, FounderProject.ProjectIndex);
+
+                var nostrPrivateKeyHex = Encoders.Hex.EncodeData(nostrPrivateKey.ToBytes());
+
+                var encryptedContent = await javascriptNostrToolsModule.InvokeAsync<string>(
```
dangershony, line 406: First make the changes I mentioned then we can review the rest of the method

src/Angor/Client/Pages/Spend.razor
```diff
@@ -57,28 +59,37 @@
...
+                        
         <div class="card mb-3">
             <div class="card-body">
                 <div class="d-flex justify-content-between">
                     <h5 class="card-title">Stage @stage.StageIndex (@stage.Stage.AmountToRelease %) - @investedAmount @network.CoinTicker (@investedCount trx) </h5>
                     @if (stageisActive)
                     {
-                        <button class="btn btn-success" disabled="@(investedCount == 0 || stage.StagePinner == true)" @onclick="() => ClaimCoinsCheckPassword(stage.StageIndex)">
+                        
+                        <button class="btn @((noCoinsToClaim) ? "btn-light" : "btn-success")" disabled="@noCoinsToClaim" @onclick="() => ClaimCoinsCheckPassword(stage.StageIndex)">
```
dangershony, line 77: `@noCoinsToClaim` excellent

src/Angor/Client/Pages/Spend.razor
```diff
@@ -183,8 +191,77 @@
                 </div>
             </div>
         </div>
+    
 
     }
+@{
```
dangershony, line 197: All this C# code should go in to the `@code {` section bellow, why have two sections? it is just confusing.

src/Angor/Client/Pages/Spend.razor
```diff
@@ -183,8 +191,77 @@
...
+    int totalStages = stageDatas.Any() ? stageDatas.Count() : 1;
+
+    // Calculate the time until the next stage release
+    var nextStage = stageDatas.Where(stage => stage.Stage.ReleaseDate > DateTime.UtcNow)
+                            .OrderBy(stage => stage.Stage.ReleaseDate)
+                            .FirstOrDefault();
+    TimeSpan? timeUntilNextStage = nextStage != null ? nextStage.Stage.ReleaseDate - DateTime.UtcNow : (TimeSpan?)null;
+}
+
+<script>
```
dangershony, line 213: why the need for the script tag?

src/Angor/Client/Pages/Spend.razor
itailiors: You are correct, i did it out of comfort to not scroll down to that section but didn't return it.
either way it seems like because it was inside the <div> so everytime you expanded it would calculate it again.
so i moved it to a function called CalculateTotalValues.
good catch

src/Angor/Client/Pages/Signatures.razor
itailiors: I forgot to mention that I did not alter the LongOperation because it comes from your ApproveSignature method. Is there a reason for it there?
or should we try to get rid of it also?

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -113,6 +113,14 @@
     bool messagesReceived;
     bool scanedForApprovals;
 
+    private bool CanApproveAllSignatures => signaturesRequests.Any(s => s.TransactionHex != null && s.AmountToInvest != null && s.TimeApproved == null);
```
dangershony, line 116: signaturesRequests could be null?

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -380,58 +388,73 @@
...
 
-        for (int i = 0; i < totalSignatures; i++)
+        private async Task ApproveAllSignatures()
         {
-            var signature = pendingSignatures[i];
+            isLoading = true;
+            StateHasChanged();
 
-            var operationResult = await notificationComponent.LongOperation(async () =>
+            var words = await passwordComponent.GetWalletAsync();
```
dangershony, line 397: I believe this wont work is password is empty, check this code to see how I did it to trigger the password prompt

https://github.com/block-core/angor/blob/main/src/Angor/Client/Pages/Invest.razor#L343

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -160,6 +174,33 @@
...
+    {
+        if (isCooldown)
+            return;
+        isLoading = true;
+        StateHasChanged();
+        isCooldown = true;
+        await FetchPendingSignatures(FounderProject);
+
+
+        if (!await EnsurePassword())
```
dangershony, line 187: What you should do is a method like this above

```
 private async Task RefreshSignaturesInternal()
    {
        
            if (passwordComponent.HasPassword())
            {
                await RefreshSignatures();
            }
            else
            {
                passwordComponent.ShowPassword(RefreshSignatures);
            }
        
    }

```

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -212,6 +253,7 @@
         }
 
         Logger.LogDebug("OnAfterRenderAsync Completed");
+        Console.WriteLine($"Signatures retrieved: {signaturesRequests.Count}");
```
dangershony, line 256: why console? use logger

src/Angor/Client/Pages/Signatures.razor
itailiors: I thought it might be code duplication, but on second thought, it significantly improves readability.
theres no need to be overly concerned with saving lines of code here. changed.

dangershony:
ok gonna look at this now

src/Angor/Client/Pages/Invest.razor
```diff
@@ -613,6 +629,9 @@
             investorProject.SignaturesInfo!.SignatureRequestEventId = investmentSigsRequest.eventId;
 
             storage.AddInvestmentProject(investorProject);
+            
+            storage.SetInvestmentAmount(investorProject.ProjectInfo.ProjectIdentifier, Investment.InvestmentAmount);
```
dangershony, line 633: was this missing? wow good catch

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -57,6 +57,11 @@
                                     <div class="col-6 d-flex align-items-center">
                                         <h6 class="mb-0">Signatures</h6>
                                     </div>
+                                    <div class="col-6 d-flex justify-content-end">
+                                        <button class="btn btn-sm @(ApproveButtonClass)" @onclick="ApproveAllSignatures" disabled="@ApproveButtonDisabled">Approve All</button>
+                                        <button class="btn btn-sm btn-secondary ml-2" @onclick="RefreshSignaturesInternal" disabled="@isCooldown">Refresh</button>
```
dangershony, line 62: I see we didn't have a refresh button, good we need one.

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -160,6 +174,33 @@
...
+            isLoading = false;
+            StateHasChanged();
+            return;
+        }
+
+        await FetchSignatures();
+        await Task.Delay(1000);
+        
+        isLoading = false;
+        isCooldown = false;
```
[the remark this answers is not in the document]
dangershony: the button becomes disabled that is normally enough but ok.

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -161,6 +175,40 @@
...
+        isLoading = true;
+        StateHasChanged();
+        isCooldown = true;
+        await FetchPendingSignatures(FounderProject);
+
+
+        await FetchSignatures();
+        await Task.Delay(1000);
+        
+        isLoading = false;
```
dangershony, line 206: Here you wanna put the bool assignments inside a `finally` look at the method `RefreshBalance` in `Wallet.razor`

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -329,25 +378,28 @@
...
             });
         }
     }
 
-    private async Task ApproveSignature(SignatureRequest signature)
+
+    private async Task<OperationResult> TryApproveSignature(SignatureRequest signature)
     {
-        var operationResult = await notificationComponent.LongOperation(async () =>
+        return await notificationComponent.LongOperation(async () =>
```
dangershony, line 389: ideally we should get rid of `LongOperation` if you feel like doing it fine otherwise in the next pr

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -357,26 +409,92 @@
...
+
+        foreach (var signature in pendingSignatures)
+        {
+            await ApproveSignatureCheckPassword(signature);
+        }
+    }
+
+    private async Task<bool> EnsurePassword()
+    {
+        if (!passwordComponent.HasPassword())
```
dangershony, line 480: interesting method

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -357,26 +409,92 @@
...
+            StateHasChanged();
+        }
     }
 
+
+
+
+        private async Task ApproveAllSignatures()
+        {
+            isLoading = true;
```
dangershony, line 450: use try/catch/finally here too, to set the bools.

src/Angor/Client/Storage/ClientStorage.cs
```diff
@@ -59,6 +59,23 @@ public void UpdateInvestmentProject(InvestorProject project)
         _storage.SetItem("projects", ret);
     }
 
+    public void SetInvestmentAmount(string projectId, decimal amount)
```
dangershony, line 62: ah I see this is a new field in storage, do we not have the inv amount specified already somewhere?

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -161,6 +175,50 @@
...
+
+        try
+        {
+            await FetchPendingSignatures(FounderProject);
+            await FetchSignatures();
+            await Task.Delay(1000);
+        }
+        catch (Exception e)
+        {
+            Console.WriteLine($"Error fetching signatures: {e.Message}");
```
dangershony, line 210: use the log class

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -161,6 +196,50 @@
...
+            });
+        }
+    }
+
+    private async Task RefreshSignatures()
+    {
+        if (isCooldown)
+            return;
+        
+        refreshButtonSpinner = true;
```
dangershony, line 219: `refreshButtonSpinner` and `isCooldown` can be combined to one variable (that worked in other paces I think it will work here too)

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -357,26 +440,101 @@
...
+        {
+            StateHasChanged();
+        }
+    }
+
+
+
+
+
+    private async Task ApproveAllSignatures()
```
dangershony, line 489: I think we have too many nested methods that we don't need, here is the flow I would suggest.

 `ApproveAllSignatures` prompt password if it is not there, then `ProcessSignatures` fetches words (so if password expires you don't bail out of the meth half way.
`ApproveSignatureCheckPassword` can go away because you already check and prompt in `ApproveAllSignatures`.
`ApproveSignature` takes the words as parameter and calls `PerformSignatureApproval` maybe even that method is a step too much) and that will show error message if sig fails.

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -357,26 +440,117 @@
...
+
+
+    private async Task ProcessSignatures()
+    {
+        var pendingSignatures = signaturesRequests.Where(s => s.TransactionHex != null && s.AmountToInvest != null && s.TimeApproved == null).ToList();
+        numOfSignatureToSign = pendingSignatures.Count;
+        numOfSignaturesSigned = 0;
+        
+        var words = await passwordComponent.GetWalletAsync();
+        if (words == null)
```
itailiors, line 531: this if can be deleted as a whole as it is a fail safe, depends how we want to treat this.

src/Angor/Client/Pages/Invest.razor
dangershony, line 633: We have the type `InvestorProject` and it has the property `InvestorProject.AmountInvested`
Why can't we use this here? why is there a difference between the model and the storage type?

I see that the property is set when we actually publish to the blockchain in this method `investorProject.CompleteProjectInvestment()`  can we not use that property instead of saving to storage a temporary page param

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -161,6 +197,49 @@
...
+            refreshButtonSpinner = false;
+            StateHasChanged();
+        }
+    }
+
+
+
     protected async Task FetchSignatures()
     {
         Logger.LogDebug("handled = {Count}, total = {SignaturesRequestsCount}", signaturesRequests.Count(x => x.AmountToInvest.HasValue), signaturesRequests.Count);
```
dangershony, line 245: Why not use string interpolation?

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -357,26 +440,117 @@
...
+                {
+                    await ProcessSignatures();
+                });
+            }
+
             passwordComponent.ClearPassword();
+        }
+        catch (Exception e)
+        {
+            Logger.LogDebug($"Error approving signatures: {e.Message}");
```
dangershony, line 513: log error not debug

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -357,26 +440,117 @@
...
+    {
+        var pendingSignatures = signaturesRequests.Where(s => s.TransactionHex != null && s.AmountToInvest != null && s.TimeApproved == null).ToList();
+        numOfSignatureToSign = pendingSignatures.Count;
+        numOfSignaturesSigned = 0;
+        
+        var words = await passwordComponent.GetWalletAsync();
+        if (words == null)
         {
-            notificationComponent.ShowErrorMessage(operationResult.Message);
+            Logger.LogDebug("words is null.");
```
dangershony, line 533: I think I rather we throw instead of doing this trick of self calling @DavidGershony what do you think?

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -357,26 +440,117 @@
...
+            {
+                await ProcessSignatures();
+            });
+            return;
+        }
+        
+        foreach (var signature in pendingSignatures)
+        {
+            await PerformSignatureApproval(signature, words);
+            numOfSignaturesSigned++;
```
dangershony, line 544: brilliant

src/Angor/Client/Pages/Invest.razor
```diff
@@ -242,9 +242,22 @@
                                 <span>Invest</span>
                             }
                         </button>
+                        
+                        @if (investorProject.SignaturesInfo != null)
+                        {
+                            <div class="mt-3">
+                                <h5>Transaction Details</h5>
+                                <div class="mb-1">
+                                <strong>Amount to invest:</strong> @storage.GetInvestmentAmount(project.ProjectInfo.ProjectIdentifier) @network.CoinTicker
```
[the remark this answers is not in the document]
itailiors: very good catch!

src/Angor/Client/Pages/Signatures.razor
itailiors: For some reason structured logging is more readable for me, changed it

src/Angor/Client/Pages/Invest.razor
```diff
@@ -573,7 +589,8 @@
                 ProjectInfo = project.ProjectInfo, 
                 Metadata = project.Metadata, 
                 SignedTransactionHex = signedTransaction!.Transaction!.ToHex(), 
-                CreationTransactionId = project.CreationTransactionId
+                CreationTransactionId = project.CreationTransactionId,
+                AmountInvested = (long) Investment.InvestmentAmount
```
dangershony, line 593: Perfect

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -357,26 +437,117 @@
...
 
-        if (operationResult.Success)
+
+
+
+
+
+    private async Task ApproveSignature(SignatureRequest signature)
+    {
+        var words = await passwordComponent.GetWalletAsync();
```
dangershony, line 460: note that the method `GetWalletAsync` can't return a null (it will throw if words are not found).
So no need to check they are not null.

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -357,26 +437,117 @@
...
+                await ProcessSignatures();
+            }
+            else
+            {
+                passwordComponent.ShowPassword(async () =>
+                {
+                    await ProcessSignatures();
+                });
+            }
+
```
dangershony, line 505: I suppose this should go in the finally block, why would we not want to reset password if we fail.

src/Angor/Client/Storage/ClientStorage.cs
```diff
@@ -59,6 +59,9 @@ public void UpdateInvestmentProject(InvestorProject project)
         _storage.SetItem("projects", ret);
     }
 
+
```
dangershony, line 62: wired that it is adding spaces for you

dangershony (approved):
Good job

PR #109, opened 2024-06-21
Storing investments list to nostr

DavidGershony:


src/Angor/Client/Pages/Invest.razor
```diff
@@ -373,9 +374,7 @@
             {
                 var words = await passwordComponent.GetWalletAsync();
 
-                var accountInfo = storage.GetAccountInfo(_networkConfiguration.GetNetwork().Name);
-
-                var nostrPrivateKey = _derivationOperations.DeriveProjectNostrPrivateKey(words, accountInfo.InvestmentsCount + 1);
+                var nostrPrivateKey = _derivationOperations.DeriveProjectNostrInvestorPrivateKey(words, project.ProjectInfo.ProjectIdentifier);
```
dangershony, line 377: We have a pattern now to ask for passwords so this will fail
check how I did it in the create
https://github.com/block-core/angor/blob/main/src/Angor/Client/Pages/Create.razor#L381

src/Angor/Client/Pages/Invest.razor
```diff
@@ -769,6 +770,37 @@
...
+            "encryptNostr",
+            rootNostrPrivateKeyHex,
+            nostrDMKey,
+            System.Text.Json.JsonSerializer.Serialize(investments));
+            
+        _RelayService.SendDirectMessagesForPubKeyAsync(rootNostrPrivateKeyHex, nostrDMKey, encryptedProjectIdList, x =>
+        {
+            if (!x.Accepted)
+            {
+                //TODO need to handle failure to save the list of projects
```
dangershony, line 799: How do we do that error message to the user?

src/Angor/Client/Pages/Investor.razor
```diff
@@ -197,6 +239,144 @@
         return input;
     }
 
+    private async Task GetProjectsAndUpdateAsync()
+    {
+        if (!passwordComponent.HasPassword())
```
dangershony, line 244: Hmm this pattern can also work actually good idea, less code.

dangershony (commented):
Other then some of my comments looks good

src/Angor/Client/Pages/Invest.razor
DavidGershony: I made the call to password component on line 361

src/Angor/Client/Pages/Invest.razor
DavidGershony: Lets have a meeting about this..

PR #111, opened 2024-06-24
Add NetworkMonitoringService to check Indexers and Relays in background 

miladsoft:


src/Angor/Client/Pages/Settings.razor
```diff
@@ -402,12 +402,28 @@
...
+
         if (!networkType.ToLower().Contains("test"))
             selectedNetwork = "mainnet";
 
         return base.OnInitializedAsync();
     }
 
+    private void UpdateUI()
+    {
+        InvokeAsync(async () =>
```
dangershony, line 415: why to await when reading from storage? we do it in many places without awaiting

src/Angor/Shared/Services/NetworkMonitoringService.cs
```diff
@@ -0,0 +1,104 @@
...
+
+    public NetworkMonitoringService(IServiceProvider serviceProvider)
+    {
+        _serviceProvider = serviceProvider;
+        _checkInterval = TimeSpan.FromMinutes(1); // Default check interval
+        _cancellationTokenSource = new CancellationTokenSource();
+    }
+
+    public void Start()
+    {
```
dangershony, line 23: If we have a timer it will constantly ping the services, right now we have some time delay on the check services that will only check if the last time was more then in interval of seconds, this is enough imo. a timer that will make http calls is unpredictable.

src/Angor/Client/Pages/Browse.razor
```diff
@@ -215,11 +216,93 @@ else
     protected override async Task OnInitializedAsync()
     {
         projects = SessionStorage.GetProjectIndexerData() ?? new List<ProjectIndexerData>();
-        if (!projects.Any())
+
+        if (projects.Any())
+        {
+            // Projects exist, initialize the search focus
+            await JS.InvokeVoidAsync("angor.initSearchFocus", "searchQuery");
```
dangershony, line 223: is there no way to focus with blazor code?
This will just put the focus on the search box right?

src/Angor/Client/Pages/Browse.razor
```diff
@@ -215,11 +216,93 @@ else
...
+            await FetchAndEnsureServices();
+        }
+    }
+
+    private async Task FetchAndEnsureServices()
+    {
+        try
         {
+            _NetworkService.AddSettingsIfNotExist();
+            await _NetworkService.CheckServices(true);
```
dangershony, line 240: here this is the right way I believe it will call the http api if last time it was called was more then x seconds

src/Angor/Client/Pages/Browse.razor
miladsoft: It is possible, but using js has better performance

src/Angor/Client/Pages/Settings.razor
miladsoft: I changed it to: 

  ```  private void UpdateUI()
    {
        settingsInfo = _clientStorage.GetSettingsInfo();
        StateHasChanged();
    }```

src/Angor/Shared/Services/NetworkMonitoringService.cs
miladsoft: I changed the code and removed the timer

src/Angor/Client/Pages/Settings.razor
dangershony: excellent

src/Angor/Shared/Services/NetworkService.cs
```diff
@@ -104,6 +106,7 @@ public async Task CheckServices(bool force = false)
                         {
                             _logger.LogError($"Failed to check indexer status url = {indexerUrl.Url}, StatusCode = {response.StatusCode}");
                         }
+                        OnStatusChanged?.Invoke();
```
dangershony, line 109: Interesting this means we don't need to wait for response it will update the page when it arrives

src/Angor/Shared/Services/NetworkMonitoringService.cs
```diff
@@ -0,0 +1,35 @@
...
+    private readonly IServiceProvider _serviceProvider;
+
+    public NetworkMonitoringService(IServiceProvider serviceProvider)
+    {
+        _serviceProvider = serviceProvider;
+    }
+
+    public async Task CheckAndEnsureServices()
+    {
+        using (var scope = _serviceProvider.CreateScope())
```
dangershony, line 17: network service is scoped? how is the scoped lifetime manages in a wasm app I wonder?

dangershony:
```
Scoped: The service is created once for the duration of the user’s session in the browser. Essentially, a scoped service in Blazor WebAssembly acts as a singleton within the context of a single user's session.
This means that once a scoped service is created, it remains the same for the entirety of the user's interaction with the Blazor WebAssembly app.
```

There is no real point in using scoped in wasm @DavidGershony

src/Angor/Shared/Services/NetworkMonitoringService.cs
```diff
@@ -0,0 +1,35 @@
...
+        {
+            var networkService = scope.ServiceProvider.GetRequiredService<INetworkService>();
+            try
+            {
+                networkService.AddSettingsIfNotExist();
+                await networkService.CheckServices(true);
+            }
+            catch (Exception ex)
+            {
+                Console.Error.WriteLine($"Error checking services: {ex.Message}");
```
dangershony, line 27: replace with logs no console

src/Angor/Shared/Angor.Shared.csproj
```diff
@@ -12,6 +12,7 @@
 
   <ItemGroup>
     <PackageReference Include="Blockcore.Core" Version="1.1.41" />
+    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="8.0.0" />
```
dangershony, line 15: why do we need `Hosting.Abstractions`? what component uses it?
does it mean we pull more data on load?

miladsoft: This is related to the previous code and should be removed

dangershony (approved):

PR #118, opened 2024-07-09
Adding a spinner on the single sig approval

dangershony:


src/Angor/Client/Pages/Signatures.razor
```diff
@@ -102,7 +102,19 @@
                                                     <td>@signature.TimeArrived.ToString("g")</td>
                                                     @if (signature.TimeApproved is null)
                                                     {
-                                                        <td><button class="btn btn-success" @onclick="() => ApproveSignatureCheckPassword(signature)">Approve</button></td>
+                                                        <td>
+                                                            <button class="btn btn-success" @onclick="() => ApproveSignatureCheckPassword(signature)" disabled="@signaturesRequestsApproving.ContainsKey(signature)">
```
itailiors, line 106: this way only the the approve signature button you pick will be disabled, maybe we should disable all signature while one is approving?

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -408,103 +423,118 @@
...
-        
-        var words = await passwordComponent.GetWalletAsync();
-        
-        foreach (var signature in pendingSignatures)
+        try
         {
-            await PerformSignatureApproval(signature, words);
-            numOfSignaturesSigned++;
-            StateHasChanged();
+            if (words == null)
```
itailiors, line 508: i'm noticing it now, words can't be null from 
`passwordComponent.GetWalletAsync();` right? thats why we removed the check before.
so we can remove the check here also

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -408,103 +423,118 @@
...
-            return new OperationResult { Success = false, Message = $"An error occurred: {ex.Message}" };
+            Logger.LogError(e, e.Message);
+            notificationComponent.ShowErrorMessage(e.Message);
+        }
+        finally
+        {
+            signaturesRequestsApproving.Remove(signature);
         }
+
+        StateHasChanged();
```
itailiors, line 452: makes sense that we want to ensure the state change only happens if an exception wasn't thrown, nice catch!

src/Angor/Client/Pages/Signatures.razor
dangershony: yes correct I will remove it

src/Angor/Client/Pages/Signatures.razor
dangershony: I believe there is a state change also inside the error  message component

itailiors (approved):

PR #123, opened 2024-07-14
Improved the UI of the project view section

miladsoft:


dangershony (approved):
Many icons added do we use all of them?

PR #130, opened 2024-07-23
Refactor UI and State Management for Checkbox and Button Handling

itailiors:


itailiors:
Updated the Claim Coins button to be enabled only if at least one checkbox for the specified stage is selected.

src/Angor/Client/Pages/Spend.razor
```diff
@@ -251,7 +251,7 @@
 
     List<(Transaction Transaction, string TrxId)> transactions = new();
 
-    Dictionary<string, string> selected = new();
+    Dictionary<int, HashSet<string>> selected = new();
```
dangershony, line 254: I see so basically here you are attempting to link a checkbox to a utxo, may I suggest using the utxo itself instead of the stage identifier and avoid the nasty hashset type.
In the `StageDataTrx` type you have both `Trxid` and `Outputindex` this is unique to the utxo, so I suggest to keep the dictionary as it was and use the `outpoint` "Trxid-Outputindex) as the key.

src/Angor/Client/Pages/Spend.razor
```diff
@@ -653,4 +688,31 @@
         await _clipboardService.WriteTextAsync(trxData);
         StateHasChanged();
     }
+    
+    public struct UtxoKey
```
dangershony, line 692: There is already such a type its called `Outpoint` use that instead

src/Angor/Client/Pages/Spend.razor
```diff
@@ -85,7 +85,7 @@
                     @if (stageisActive)
                     {
 
-                        <button class="btn @((noCoinsToClaim) ? "btn-light" : "btn-success")" disabled="@noCoinsToClaim" @onclick="() => ClaimCoinsCheckPassword(stage.StageIndex)">
+                        <button class="btn @((noCoinsToClaim) ? "btn-light" : "btn-success")" disabled="@((noCoinsToClaim || !IsCheckboxSelectedForStage(stage.StageIndex)))" @onclick="() => ClaimCoinsCheckPassword(stage.StageIndex)">
```
dangershony, line 88: This can be much simplified, when we expand another stage we clear the dictionary so there will only be items for a particular stage right? so you just need to check if the dictionary `selectedUtxos` has items in it, correct me if I am wrong.

src/Angor/Client/Pages/Spend.razor
```diff
@@ -628,20 +644,39 @@
...
-        if (selected.ContainsKey(trxId))
+        if (selectedUtxos.ContainsKey(key))
         {
-            selected.Remove(trxId);
+            selectedUtxos.Remove(key);
         }
         else
         {
-            selected.Add(trxId, null);
+            selectedUtxos[key] = true;
```
dangershony, line 658: This will always be true, we wont have a case that the value is false right?

src/Angor/Client/Pages/Spend.razor
```diff
@@ -653,4 +688,31 @@
...
+        public override bool Equals(object obj)
+        {
+            if (!(obj is UtxoKey))
+                return false;
+
+            var key = (UtxoKey)obj;
+            return Trxid == key.Trxid && Outputindex == key.Outputindex;
+        }
+
+        public override int GetHashCode()
```
dangershony, line 710: ah I see you need to compile a hashcode for the dictionary unique key.
instead of using a new type it would be simpler to use a string of "trxid-index", but keep this as it is.

src/Angor/Client/Pages/Spend.razor
itailiors: `noCoinsToClaim` is determined by `investedCount == 0` or `stage.StagePinner == true`. 
 there might be a scenario where `selectedUtxos` has values for stage 2. In this case, the `claimCoins` option for stage 1 won't be disabled. 
 providing the `stageIndex` and the relevant method, we can handle the logic for each stage appropriately.

src/Angor/Client/Pages/Spend.razor
itailiors: removing the key means it no longer exists in the dictionary, so there isn't a false value associated with it?
if so then i think you're right

src/Angor/Client/Pages/Spend.razor
itailiors: Will check it out 👍

PR #135, opened 2024-08-07
Add display option to see btc amount in fiat 

itailiors:


src/Angor/Client/Angor.Client.csproj
```diff
@@ -15,7 +15,10 @@
     <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly.DevServer" Version="8.0.7" PrivateAssets="all" />
     <PackageReference Include="NBitcoin" Version="7.0.25" />
     <PackageReference Include="Nostr.Client" Version="2.0.0" />
+    <PackageReference Include="System.Text.Json" Version="8.0.4" />
     <PackageReference Include="QRCoder" Version="1.6.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly" Version="8.0.7" />
+    <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly.DevServer" Version="8.0.7" PrivateAssets="all" />
```
itailiors, line 21: note: this wasn't here before merge.

src/Angor/Client/Services/CurrencyService.cs
```diff
@@ -0,0 +1,108 @@
...
+        // Check if we have a recent rate in the cache
+        if (_rateCache.ContainsKey(preferredCurrency) && 
+            DateTime.UtcNow - _rateCache[preferredCurrency].timestamp < _cacheDuration)
+        {
+            _logger.LogInformation($"Using cached rate for {preferredCurrency}");
+            return _rateCache[preferredCurrency].rate;
+        }
+
+        // Fetch new rate if not cached or cache is stale
+        string apiUrl = "https://mempool.space/api/v1/prices";
```
dangershony, line 57: The part that fetches should move to its own service

itailiors: I've moved it to its own service, still we need to figure out with mempool with our use is alright.
right now i made it that for each instance of angor, it will request the rates. (and will cache for 10 minutes).
still I am not sure if its requesting to much times from mempool

itailiors:
still in prototype for the websocket!
had to use server because CORS issues when directly connecting to external WebSocket APIs from the client.
still there is a problem (unrelated to websocket) that i need to fix, and to change it that the rate doesnt fetch everysecond.
(and add/check support for other websockets except binance)

itailiors:
also for some reason there are still the spaces, but now i am not sure if they are good because i added rider's auto reformat on save so it might be that 🤔 💭

src/Angor/Client/Pages/Wallet.razor
```diff
@@ -1255,4 +1260,25 @@ else
...
+    private async Task PrecomputeFiatValues()
+    {
+        fiatValues.Clear();
+        var addresses = accountBalanceInfo.AccountInfo.AllAddresses();
+
+
+        foreach (var addressInfo in addresses)
+        {
+            var total = Money.Satoshis(addressInfo.Balance).ToUnit(MoneyUnit.BTC);
+            var fiatValue = await _currencyService.GetBtcValueInPreferredCurrency(total); // Calls the method from ICurrencyService
```
DavidGershony, line 1279: Can't you just get the BTC value for 1 fiat and calculate the rest locally? it seems like an overkill to call the server every time for a simple multiplication

itailiors: Yeah still have i little redesign to do from when its changed to a websocket api,
generally it's supposed to be cached for 10 minutes and then if its not cached to pull it back again,
now with the websocket always having the value a different solution might be appropriate

src/Angor/Client/Services/CurrencyService.cs
```diff
@@ -0,0 +1,98 @@
...
+        {
+            // Log initial state
+            Console.WriteLine($"[DEBUG] Fetching BTC value for balance: {btcBalance}");
+
+            // Get the user's preferred currency code from storage
+            var currencyCode = _storage.GetCurrencyDisplaySetting();
+            Console.WriteLine($"[DEBUG] Preferred currency code: {currencyCode}");
+
+            // Fetch the conversion rate from the rate service
+            var rate = await _currencyRateService.GetBtcToCurrencyRate(currencyCode);
```
DavidGershony, line 35: I think we should cache the result here and use it for a fixed time frame

PR #147, opened 2024-09-04
Btc to fiat rest api

itailiors:


itailiors:
Generally finished,
I have these questions for this.
1) @DavidGershony we talked about that currencyservice.cs can go into balanceDisplay but the current architectural of this doesn't work. (if we look at the utxo each of them is its own component they aren't the same one).
is there a better way to design it then? (right now i can think only that it can be precomputed like it is in the PR).

2) i added `if (_NetworkConfiguration.GetNetwork().CoinTicker.Equals("BTC", StringComparison.OrdinalIgnoreCase))` because tbtc isnt real money (david's comment), so right now no users will see the actual change so we won't get feedback if theres something wrong.
should i comment it out for now?

3) there are more places I Could put balancedisplay but i'll wait to know if the logic is correct (because of 1)

miladsoft:
> Generally finished, I have these questions for this.
> 
> 1. @DavidGershony we talked about that currencyservice.cs can go into balanceDisplay but the current architectural of this doesn't work. (if we look at the utxo each of them is its own component they aren't the same one).
>    is there a better way to design it then? (right now i can think only that it can be precomputed like it is in the PR).
> 2. i added `if (_NetworkConfiguration.GetNetwork().CoinTicker.Equals("BTC", StringComparison.OrdinalIgnoreCase))` because tbtc isnt real money (david's comment), so right now no users will see the actual change so we won't get feedback if theres something wrong.
>    should i comment it out for now?
> 3. there are more places I Could put balancedisplay but i'll wait to know if the logic is correct (because of 1)

I believe it's sufficient to display the price only in the wallet and specifically for Bitcoin. I initially planned to implement this on the main page, but considering Angor's primary mission, I see this as an unnecessary addition.

src/Angor/Client/Angor.Client.csproj
```diff
@@ -15,7 +15,10 @@
     <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly.DevServer" Version="8.0.7" PrivateAssets="all" />
     <PackageReference Include="NBitcoin" Version="7.0.25" />
     <PackageReference Include="Nostr.Client" Version="2.0.0" />
+    <PackageReference Include="System.Text.Json" Version="8.0.4" />
     <PackageReference Include="QRCoder" Version="1.6.0" />
+    <PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly" Version="8.0.8" />
```
dangershony, line 20: any idea what this references are? do we need them still?

src/Angor/Client/Pages/Index.razor
```diff
@@ -20,10 +21,33 @@
...
             </div>
         </div>
     </div>
 </div>
 
 
+@code{
+    private string btcBalanceInUsd;
+ 
+    protected override async Task OnInitializedAsync()
```
dangershony, line 34: so wait we fetch balance only once on startup?

src/Angor/Client/Services/CurrencyRateService.cs
```diff
@@ -0,0 +1,98 @@
...
+
+    public async Task<decimal> GetBtcToCurrencyRate(string currencyCode)
+    {
+        if (string.IsNullOrWhiteSpace(currencyCode))
+            throw new ArgumentException("Currency code must not be null or empty.", nameof(currencyCode));
+
+        var preferredCurrency = currencyCode.ToUpper();
+
+        // Check if the rate is cached and still valid
+        var cachedRate = GetCachedRate(preferredCurrency);
```
dangershony, line 28: ah I see we have caching ok

src/Angor/Client/Storage/LocalSessionStorage.cs
```diff
@@ -42,26 +37,6 @@ public bool IsProjectMetadataStorageByPubkey(string pubkey)
         return _sessionStorageService.GetItem<Project>(projectId);
     }
 
-    public void AddProjectIdToNostrPubKeyMapping(string npub, string projectId)
```
dangershony, line 45: was this not used?

src/Angor/Client/Storage/LocalSessionStorage.cs
```diff
@@ -96,19 +71,66 @@ public void SetUnconfirmedOutboundFunds(List<Outpoint> unconfirmedInfo)
...
     {
         _sessionStorageService.Clear();
     }
+
+    public void StoreProjectMetadata(string pubkey, ProjectMetadata projectMetadata)
+    {
+        _sessionStorageService.SetItem(pubkey, projectMetadata);
+    }
+
+    public void AddProjectIdToNostrPubKeyMapping(string npub, string projectId)
```
dangershony, line 112: ah sorry it moved down ok

src/Angor/Client/Pages/Wallet.razor
```diff
@@ -16,23 +7,28 @@
...
 @inject IClipboardService _clipboardService
+@inject ICurrencyService _currencyService
 @inject IDerivationOperations _derivationOperations
 @inject NavMenuState NavMenuState
 @inject IEncryptionService _encryptionService
 @inject IClipboardService ClipboardService
-
-
+@using System.Text.Json
+@using Angor.Client.Storage
```
dangershony, line 16: why move the `using` under the `injected` types? out convention is that they are above, please fix this.

src/Angor/Client/Pages/Wallet.razor
```diff
@@ -1255,4 +1257,48 @@ else
...
+
+        var addresses = accountBalanceInfo.AccountInfo.AllAddresses().ToList();
+        
+        if (addresses.Count == 0)
+        {
+            Logger.LogWarning("No addresses available for processing.");
+            return;
+        }
+
+        var btcBalances = addresses
```
dangershony, line 1285: why not use the already existing type `accountBalanceInfo.TotalBalance` ?

dangershony (changes requested):
The tow last comments we should change before we push

dangershony:
> considering Angor's primary mission, I see this as an unnecessary addition.

I agree, good call

src/Angor/Client/Pages/Index.razor
miladsoft: removed from index

src/Angor/Client/Pages/Wallet.razor
miladsoft: fixes

src/Angor/Client/Pages/Wallet.razor
miladsoft: Use GetBtcValuesInPreferredCurrency for fiat conversion.

dangershony (approved):

src/Angor/Client/Angor.Client.csproj
dangershony: what about this?

dangershony (approved):

PR #168, opened 2024-11-06
Add RawTransactionModal for TRX processes 

itailiors:
added RawTransactionModal as a component,
add for classes spend,invest,wallet.

added max money to invest (5 btc)

### DIFF
--- a/src/Angor/Client/Pages/Invest.razor
+++ b/src/Angor/Client/Pages/Invest.razor
@@ -12,6 +12,8 @@
 @using Nostr.Client.Keys
 @using System.Diagnostics
 @using Angor.Client.Models
+@using System.Text.Json
+
 
 @inherits BaseComponent
 
@@ -242,6 +244,11 @@
                             }
                         </button>
                     </div>
+                    <button class="btn btn-primary" @onclick="ShowTransactionJsonModal">
+                        Show Transaction Details
+                    </button>
+
+                    <RawTransactionModal RawTransactionJson="@transactionJson" IsVisible="@showTransactionJsonModal" IsVisibleChanged="HandleTransactionJsonModalVisibility"/>
                 </div>
             </div>
         </div>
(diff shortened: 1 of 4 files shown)


itailiors:
Hey @dangershony, I just noticed your note in the task:

"We should limit this on the invest page.
The cap should be 0.1 bitcoin?"

Currently, we’ve set a minimum of 2 BTC on testnet 😅

It seems like we have a couple of options (or let me know if you think of a smarter approach):

1) Add fields during project creation for setting both minimum and maximum investments. This way, we could set a cap at 0.1 BTC. However, this might not be ideal if founders later need to adjust these values.

2) Separate logic for TBTC and BTC on the invest page, though this could lead to hardcoding.


Option 1 (with some variation) seems like the better one, allowing founders to define their own minimum and maximum.

(in the meanwhile i'll just set the minimum for 0.01 and max for 0.1 :) )

dangershony:
> Separate logic for TBTC and BTC on the invest page, though this could lead to hardcoding.

I think that's fine, no limit on testnet, on mainnet we put a limit.
just use the network type property 

```
        var network = _networkConfiguration.GetNetwork();
        if (network.NetworkType == NetworkType.Mainnet)
        {
            // limit amount
        }

```

src/Angor/Client/Pages/Invest.razor
```diff
@@ -531,12 +539,19 @@ else
     {
         if (Investment.IsSeeder)
         {
-            var minSeederAmount = 2;
+            var minSeederAmount = 0.01m;
```
dangershony, line 542: Seeder is not implemented yet (the new name is lead investor) so don't do anything here it will just be confusing)

src/Angor/Client/Pages/Invest.razor
```diff
@@ -531,12 +539,19 @@ else
...
+            var maxSeederAmount = 0.1m;
             if (Investment.InvestmentAmount < minSeederAmount)
             {
                 notificationComponent.ShowErrorMessage($"Seeder minimum investment amount of {minSeederAmount} BTC was not reached");
                 return;
             }
+            
+            if (Investment.InvestmentAmount > maxSeederAmount)
```
dangershony, line 550: `if(network.NetworkType == NetworkType.Mainnet && Investment.InvestmentAmount > maxInvestAmount)`

dangershony (approved):
Looks good apart for the limitation on the lead investor and checking if it is mainnet
Fix that in a new PR

PR #169, opened 2024-11-07
Add cleanup on rescan

itailiors:
@dangershony @DavidGershony 
After removing the change from [this commit](https://github.com/block-core/angor/pull/136/commits/660765ee662d1f3f91093c34f61a8fc5076e237a), 
I noticed an issue during rescanning: if another project is found, duplicate projects can appear. 

If there's a more effective solution that avoids deleting i'll implement that

p.s, Maybe we can do a softer filter in LookupProjectKeysOnIndexerAsync?

image

### DIFF
--- a/src/Angor/Client/Pages/Founder.razor
+++ b/src/Angor/Client/Pages/Founder.razor
@@ -34,12 +34,15 @@
 
                 </div>
             </div>
-            <button class="btn btn-border" @onclick="NavigateToCreateProject">
+            <button class="btn btn-border" @onclick="NavigateToCreateProject" disabled="@scanningForProjects">
                 <i>
                     <Icon IconName="add"></Icon>
                 </i>
-                <span class="nav-link-text ms-1">Create Project</span>
+                    <span class="nav-link-text ms-1">
+                    @(scanningForProjects ? "Scanning..." : "Create Project")
+                </span>
             </button>
+
         </div>
         <p class="mb-0 font-weight-normal text-sm mt-4">
             To create a new project or view your existing projects, an on-chain transaction and a Nostr DID are required.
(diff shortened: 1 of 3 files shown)


src/Angor/Client/Pages/Founder.razor
```diff
@@ -107,6 +107,9 @@
     private async Task LookupProjectKeysOnIndexerAsync()
     {
         scanningForProjects = true;
+        // Clear the existing projects before rescan
+        founderProjects.Clear();
```
dangershony, line 111: This is very odd because line 120 should prevent those duplicates

dangershony: also line 146 can be a place to look at why we get duplicates

dangershony:
Any update on this PR?

itailiors:
> Any update on this PR?

I tried to have a "softer" limiter but it still is the same, so when i have time i'll debug this to see exactly what needs to be excluded

itailiors:
> Any update on this PR?

ready to check, 

the logic now checks if a project already exists:

if found: updates metadata or project info.
if not: adds it as new.
no clearing, just merging and syncing with storage. avoids duplicates efficiently.

src/Angor/Client/Pages/Founder.razor
```diff
@@ -109,63 +112,104 @@
...
+                        }
+                        else
+                        {
+                            var projectIndex = founderProjectsToLookup.Keys.ToList().IndexOf(e.Pubkey) + 1;
+                            var trxId = founderProjectsToLookup[e.Pubkey].TrxId;
+
+                            founderProjects.Add(new FounderProject
+                            {
+                                ProjectInfo = projectInfo,
+                                ProjectIndex = projectIndex,
```
dangershony, line 177: what is the `ProjectIndex` is it not the index of the project in the derivation path?

dangershony (commented):
It looks fine to me but @DavidGershony should also have a look

dangershony (approved):

PR #175, opened 2024-11-18
Project creation refactor 

itailiors:
opening this PR instead for it to be a draft while its in testing


dangershony:
So can I test this now?

dangershony:
@miladsoft will have to give it a touch once it is pushed

image

dangershony:
Ideally the predefined stages will be a popup.
but I think we can do that in a separate PR, if this one is good enough I would rather push it already and make a popup for predefined stages separately

src/Angor/Client/Pages/Create.razor
```diff
@@ -204,293 +189,425 @@
...
+                    <span>Create Project...</span>
+                }
+                else
+                {
+                    <span>Create Project</span>
+                }
+            </button>
+
+            <button class="btn btn-danger ms-3"
+                    @onclick="@(() => ShowResetConfirmation("nostrMetadataCreated"))"
```
dangershony, line 422: Reset info doesn't seem to work, I think it should move you back to the info page
Also it is not clear what button resets the metadata, do we need the reset buttons? a user can just go back and do it manually on the tab they need, it might be too confusing IMO.

dangershony:
I wonder if this is not a bit confusing, can the user do anything after creating the profile on Nostr?
We should at least disable all the other buttons, they get disabled if I navigate away and then back but not if I navigate to the previous tab

image

dangershony:
This is not so clear, bug I guess its is more a UI change 

image

src/Angor/Client/Pages/Create.razor
itailiors: once you create a nostr metada, or project info (which is UI only), you are restricted from changing them.

mainly because of the nostr one that once its created it's on nostr protocol.

i added these buttons as a "fail safe", incase there's an error that isn't from nostr, or because of something unexpected. should we do it in another way?

itailiors:
> I wonder if this is not a bit confusing, can the user do anything after creating the profile on Nostr? We should at least disable all the other buttons, they get disabled if I navigate away and then back but not if I navigate to the previous tab
> 
> image

after a user creates the nostr meta data, he needs to fill in the project info and create that.
he cannot create the project yet.

although we can add that the active tab is sent to where it has been reset, but the main reason of this PR is to give the user the option to do anything he wants to do here, and finalize his project when he wants. (Depending on what we do on the fail safe i'll adjust this)

itailiors:
> This is not so clear, bug I guess its is more a UI change


you're correct, i'll try to make it more apparent that you have 2 choices to create stages.

src/Angor/Client/Pages/Create.razor
dangershony: yes I think we don't need to have this reset, once the creator posts the metadata to nostr then we must disable every thing else and only enable the button to publish to the bitcoin network.

A user can always just navigate backwards and make the changes they want anyway.
The reset just makes it confusing.

src/Angor/Client/Pages/View.razor
```diff
@@ -443,107 +451,135 @@
...
+            }
 
-                if (project == null)
-                {
-                    findInProgress = true;
+            project = SessionStorage.GetProjectById(ProjectId);
+            if (project != null)
+            {
+                projectStats.Loading = false;
+                StateHasChanged();
```
dangershony, line 506: Shouldn't you be calling the method `SetProjectLinksAndRefreshBalance` here?
I see that the explorer link is missing

src/Angor/Client/Pages/Create.razor
```diff
@@ -578,45 +704,49 @@
...
-        await Task.Delay(10);
+        confirmationMessage = "Are you sure you want to finalize the Nostr Profile creation? Once done, changes can only be made on Nostr.";
+        finalizeAction = FinalizeNostrProfileCreation;
+        showConfirmationModal = true;
+    }
 
-        try
+    private async void ConfirmFinalizeProjectInfoCreation()
+    {
+        if (!await ProjectInfoMetaDataTestInput())
```
dangershony, line 718: This should obviously be called before we publish to Nostr probably in the method `CreatNostrProfile` or even in `FinalizeNostrProfileCreation` otherwise we will post the project info to Nostr and then find it fails validation

dangershony:
I realise we have 3 operations that need to happen before a project is finalized
- post project metadata to nostr
- post project info to nostr
- post a trx to the blockchain

We should probably have 3 buttons that each becomes enabled at their turn (I think it is like that now?)
If any of the stages failed the use should be able to try again that stage until is succeeds

dangershony:
I noticed when I create a project we have two entries in the database for that project one is missing the trx id

image

src/Angor/Client/Pages/Create.razor
```diff
@@ -851,6 +989,32 @@
...
+            // is this the correct way to get the address @dangershony @davidgershony
+            var changeAddress = accountInfo.ChangeAddressesInfo.FirstOrDefault(addr => !addr.UtxoData.Any())?.Address;
+            if (string.IsNullOrEmpty(changeAddress))
+            {
+                notificationComponent.ShowErrorMessage("No available change address found.");
+                return;
+            }
+
+            // reconstruct the transaction using refreshed UTXO data
+            signedTransaction = _WalletOperations.AddInputsAndSignTransaction(
```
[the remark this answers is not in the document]
itailiors: now if a project already exists using the projectidentifier before creating or initializing it. if it exists, the project is updated directly in place instead of creating a new one. this ensures that all updates, including the assignment of creationtransactionid, modify the same project entry throughout the workflow.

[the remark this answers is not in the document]
itailiors: also, does this have the required outputs? 
https://explorer.angor.io/tbtc/explorer/transaction/04273e7eb6be1d9486d76dc210de92b764544e6c597f470a0ca8735612f78925

src/Angor/Client/Pages/Create.razor
itailiors: how is it possible we will post not validated project info?
Right now the order is after we create nostr profile, in the method we:
1) validate the inputs `if (!await ProjectInfoMetaDataTestInput())
        {
            // If validation fails, return early
            return;
        }`
        
        
2) post it to the relay with AddProjectAsync

(or when you say validation you mean other type of validation?)

src/Angor/Client/Pages/Create.razor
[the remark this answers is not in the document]
dangershony: > also, does this have the required outputs?  
Yes this looks correct

src/Angor/Client/Pages/Create.razor
```diff
@@ -84,116 +83,102 @@
...
 
 
-
-
 <div class="card card-body mt-4">
-    @if (activeTab == 1)
-    {
-        <div class="row g-4">
-            <div class="col-lg-7 col-12">
+@if (activeTab == 1)
```
dangershony, line 87: the if block should be pushed more to the right

src/Angor/Client/Pages/Create.razor
```diff
@@ -204,352 +189,598 @@
...
-                <label for="penaltyDays" class="form-label">Penalty Days</label>
-                <InputNumber id="penaltyDays" @bind-Value="project.ProjectInfo.PenaltyDays" class="form-control" />
-            </div>
+        <!-- Presets for Stages -->
+        <!-- Toggle Button -->
+        <div class="d-flex justify-content-between align-items-center mb-3">
+            <button class="btn btn-primary" @onclick="ToggleStageMode">
+                @if (isPresetMode)
+                {
+                <span>Switch to Automatic</span>
```
dangershony, line 258: Perhaps "switch to manual helper" is more correct?

src/Angor/Client/Pages/Create.razor
```diff
@@ -204,352 +189,598 @@
...
 
-            <!-- Expiry Date -->
-            <div class="mb-3">
-                <label for="expiryDate" class="form-label">Expiry Date</label>
-                <InputDate id="expiryDate" @bind-Value="project.ProjectInfo.ExpiryDate" class="form-control" />
-            </div>
+        <!-- Conditionally Render Presets or Automatic Inputs -->
+        @if (isPresetMode)
+        {
+        <!-- Preset Selection -->
```
dangershony, line 270: notice the html blocks are not aligned they need to be pushed right

src/Angor/Client/Pages/Create.razor
```diff
@@ -204,352 +189,598 @@
...
-                project.ProjectIndex = projectsKeys.Index;
-                project.ProjectInfo.FounderKey = projectsKeys.FounderKey;
-                project.ProjectInfo.FounderRecoveryKey = projectsKeys.FounderRecoveryKey;
-                project.ProjectInfo.ProjectIdentifier = projectsKeys.ProjectIdentifier;
-                project.ProjectInfo.NostrPubKey = projectsKeys.NostrPubKey;
-            }
+                ProjectIndex = nextProjectIndex,
+                Metadata = new ProjectMetadata(),
+                ProjectInfo = new ProjectInfo
+                {
```
dangershony, line 641: I prefer that we don't fill the stage now that we have presets.

src/Angor/Client/Storage/ClientStorage.cs
```diff
@@ -112,83 +101,66 @@ public List<FounderProject> GetFounderProjects()
 
     public void UpdateFounderProject(FounderProject project)
     {
-        var projects = _storage.GetItem<List<FounderProject>>("founder-projects");
-        
-        var item = projects.FirstOrDefault(f => f.ProjectInfo.ProjectIdentifier == project.ProjectInfo.ProjectIdentifier);
+        var projects = _storage.GetItem<List<FounderProject>>("founder-projects") ?? new List<FounderProject>();
+
+        var existingIndex = projects.FindIndex(p =>
+            p.ProjectInfo.ProjectIdentifier == project.ProjectInfo.ProjectIdentifier &&
```
dangershony, line 107: no this is wrong, ProjectIdentifier must be unique if we have more then one it means we are incorrectly using the index else where

dangershony:
I was able to add a stage after creating the profile 

image

now I can not remove the stage the remove button seems to have disappeared.

src/Angor/Client/Pages/Create.razor
```diff
@@ -204,352 +189,598 @@
...
-            <h3 class="mb-3">Stages</h3>
-            @foreach (var stage in project.ProjectInfo.Stages)
+            @if (project.ProjectInfo.Stages.IndexOf(stage) > 0)
             {
-                <div class="mb-3">
-                    <label class="form-label">The amount in % to allocate for stage @(project.ProjectInfo.Stages.IndexOf(stage) + 1)</label>
-                    <div class="input-group">
-                        <InputNumber @bind-Value="stage.AmountToRelease" class="form-control" placeholder="Enter amount to release as a percentage" min="0.01" max="100" step="0.01" />
-                        <InputDate @bind-Value="stage.ReleaseDate" class="form-control" />
-                        <button type="button" class="btn btn-danger" @onclick="() => RemoveStage(stage)">Remove</button>
```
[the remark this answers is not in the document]
dangershony: ah I think I realize what happened, if I refresh the page then the buttons are disabled correctly.
I think the page is not refreshed or the items on the page dont reflect the items in the database.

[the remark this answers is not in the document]
dangershony: Also after I create the profile and refresh the page then the create project info button is now disabled

dangershony:
The create button should be like the other two in my opinion
it looks nice like that
image

src/Angor/Client/Pages/Create.razor
```diff
@@ -591,30 +804,29 @@
...
                 _ =>
                 {
                     createProfileSpinner = false;
 
                     if (!_.Accepted)
-                        notificationComponent.ShowErrorMessage("Failed to store the project information on the relay!!!"); //TODO add export project info
-
-                    if (nostrMetadataCreated == false)
+                    {
+                        notificationComponent.ShowErrorMessage("Failed to store the project information on the relay!");
```
dangershony, line 816: Lets start to log such errors as well? (perhaps inside the method `ShowErrorMessage`) I realise we did not do it before enough.

src/Angor/Client/Pages/Create.razor
```diff
@@ -204,352 +189,598 @@
...
-            var projectsKeys = _derivationOperations.GetProjectKey(keys, latestProject?.ProjectIndex + 1 ?? 1);
+        if (projects.Count(p => !string.IsNullOrEmpty(p.CreationTransactionId)) >= keys.Keys.Count)
+        {
+            notificationComponent.ShowErrorMessage("All founder keys have been used for this wallet!");
+            return;
+        }
 
-            var fromStorage = projects.FirstOrDefault(p => p.ProjectInfo.ProjectIdentifier == projectsKeys.ProjectIdentifier);
+        var latestProject = projects
+            .Where(p => !string.IsNullOrEmpty(p.CreationTransactionId))
```
dangershony, line 619: I in this new approach we do not push the project information to storage until we call the method created profile button.
It means you cannot navigate away and come back to continue editing the project.

Perhaps we should push to storage the new project immediately now and later update it as we create metadata
For example when we navigate away from each tab we could save the project to storage.
Until the create profile was called then we do not modify anymore the project class (we must also make sure we save to storage the fact the profile was created (not sure how we do it right now)

src/Angor/Client/Pages/Create.razor
dangershony: My point is we must do all validation before we create the Nostr profile, if not it is too confusing to the user imo
Imagine the metadata is not valid so they create a profile then have to go back to fix the metadata? it is also harder to check that logic programmatically.

dangershony: so basically I suggest make a new method something like this before creating the profile

```
 private async void Validate()
 {
     if (!await ValidateProfile())
     {
         // If validation fails return
         return;
     }

     if (!await ValidateMetadata())
     {
         // If validation fails return
         return;
     }
    
   // continue execution to create profile
 }
```

src/Angor/Client/Pages/Create.razor
```diff
@@ -782,19 +916,31 @@
...
+            signedTransaction = _WalletOperations.AddInputsAndSignTransaction(
+                accountInfo.GetNextChangeReceiveAddress(),
+                unsignedTransaction,
+                words,
+                accountInfo,
+                feeData.SelectedFeeEstimation
+            );
 
-            signedTransaction = _WalletOperations.AddInputsAndSignTransaction(accountInfo.GetNextChangeReceiveAddress(), unsignedTransaction, words, accountInfo, feeData.SelectedFeeEstimation);
+            project.CreationTransactionId = signedTransaction.Transaction.GetHash().ToString();
```
dangershony, line 942: I don't think it is correct to set `CreationTransactionId` and update storage here, because we did not publish to the blockchain yet, notice we set those values later when we publish (in the method `send`) if we don't get back an error from the indexer.

src/Angor/Client/Pages/Create.razor
```diff
@@ -893,4 +1048,201 @@
...
+    }
+    
+
+    private void ToggleStageMode()
+    {
+        isPresetMode = !isPresetMode; // Toggle between Presets and Automatic
+        StateHasChanged(); // Update the UI
+    }
+    
+    private bool ValidateTransactionOutputs(Transaction transaction)
```
dangershony, line 1228: this method seems to not be used at all?

src/Angor/Client/Pages/Create.razor
itailiors: do you mean missing name? or some of the metadata?
we have some checks in ConfirmFinalizeNostrProfileCreation before the pop up of creating

dangershony: I mean the point in time for the validation of `ConfirmFinalizeNostrProfileCreation` is too late, it happens after the profile was already created on nostr.

itailiors:
@dangershony for each different indexer it seems like the api route is different, for example mempool doesnt have "query" like angor indexer.
how do we want to address that?
possible solution is to have route normalization or api mapping.
also we can just `not support` it (even though it might be on the same network)

links for genesis block hashes - 
angor testnet indexer link  : https://tbtc.indexer.angor.io/api/query/block/index/0
mempool testnet indexer link:  https://mempool.space/testnet/api/block-height/0
mempool signet : https://mempool.space/signet/api/block-height/0

dangershony:
There are conflicts, can you fix them before I dig in to this?

src/Angor/Client/Pages/View.razor
```diff
@@ -443,60 +451,70 @@
...
-                    if (projectIndexerData != null)
-                    {
-                        project = new Project { CreationTransactionId = projectIndexerData.TrxId };
-                       
-                        _RelayService.LookupProjectsInfoByEventIds<NostrEvent>(x =>
+            if (projectIndexerData != null)
+            {
+                project = new Project { CreationTransactionId = projectIndexerData.TrxId };
+
+                _RelayService.LookupProjectsInfoByEventIds<NostrEvent>(x =>
```
itailiors, line 517: @dangershony @DavidGershony on this relay call i just took what's in main and relaying on it

itailiors:
> There are conflicts, can you fix them before I dig in to this?

merged

PR #179, opened 2024-11-21
Check block when adding indxer

itailiors:


src/Angor/Shared/Services/IndexerService.cs
```diff
@@ -225,5 +237,50 @@ public async Task<string> GetTransactionHexByIdAsync(string transactionId)
...
+                response.EnsureSuccessStatusCode();
+                
+                var responseContent = await response.Content.ReadAsStringAsync();
+                return responseContent.Trim();
+
+
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"Error fetching genesis block hash: {ex.Message}");
```
dangershony, line 255: use log framework not console

src/Angor/Shared/Services/IndexerService.cs
```diff
@@ -225,5 +237,50 @@ public async Task<string> GetTransactionHexByIdAsync(string transactionId)
...
+            // Compare the first 64 characters (32 bytes in hex) of the fetched hash to the expected hash
+            return fetchedHash.StartsWith(expectedHash, StringComparison.OrdinalIgnoreCase);
+        }
+
+
+        public async Task<bool> CheckIndexerStatusAsync(string indexerUrl)
+        {
+            try
+            {
+                using var httpClient = new HttpClient();
```
dangershony, line 271: I believe we must inject the http client not create one

src/Angor/Client/NetworkConfiguration.cs
```diff
@@ -198,5 +198,19 @@ public List<SettingsUrl> GetDefaultExplorerUrl()
...
+    public string GetGenesisBlockHash()
+    {
+        // Determine the correct genesis block hash based on the network type
+        return currentNetwork.NetworkType.ToString() switch
+        {
+            "Mainnet" => "000000000019d6689c085ae165831e93",
+            "Testnet" => "000000000933ea01ad0ee984209779ba",
+            "Signet" => "00000000020f01e33f91b6c7c5d3a3f8",
+            "Regtest" => "0f9195cbdb894feda6ee07798e0d597d",
+            _ => throw new NotSupportedException($"Network type {currentNetwork.NetworkType.ToString()} is not supported")
```
dangershony, line 211: what about angornet?

src/Angor/Client/Pages/Settings.razor
```diff
@@ -471,39 +472,56 @@
...
         if (!string.IsNullOrWhiteSpace(newIndexerLink))
         {
             if (!Uri.TryCreate(newIndexerLink, UriKind.Absolute, out Uri? uri))
             {
-                notificationComponent.ShowErrorMessage($"invalid url {newIndexerLink}");
+                notificationComponent.ShowErrorMessage($"Invalid URL: {newIndexerLink}");
                 return;
             }
 
-            if (uri.Scheme is not ("http" or "https"))
```
dangershony, line 485: why remove the schema test? cant remember why we put it there, probably to not confuse with relays

src/Angor/Client/Pages/Settings.razor
```diff
@@ -471,39 +472,56 @@
...
+
+            if (settingsInfo.Indexers.Any(a => a.Url == newIndexerLink))
             {
-                notificationComponent.ShowErrorMessage($"invalid url {newIndexerLink} schema must be http or https");
+                notificationComponent.ShowErrorMessage($"URL already exists: {newIndexerLink}");
                 return;
             }
 
-            newIndexerLink = new Uri($"{uri.Scheme}://{uri.Host}").AbsoluteUri.TrimEnd('/');
+            // Temporarily skip online check
```
dangershony, line 496: Why the skip?

src/Angor/Client/Pages/Settings.razor
```diff
@@ -471,39 +472,56 @@
...
+            // }
 
-            if (settingsInfo.Indexers.Any(a => a.Url == newIndexerLink))
+            var fetchedHash = await _indexerService.GetGenesisBlockHashAsync(newIndexerLink);
+            var expectedHash = _networkConfiguration.GetGenesisBlockHash();
+
+                if (string.IsNullOrEmpty(fetchedHash) || !_indexerService.ValidateGenesisBlockHash(fetchedHash, expectedHash))
             {
-                notificationComponent.ShowErrorMessage($"url exists {newIndexerLink}");
+                notificationComponent.ShowErrorMessage("The indexer is not compatible with the selected network.");
```
dangershony, line 509: Perhaps also write expected network name and what was provided (if any)

src/Angor/Shared/Services/IndexerService.cs
```diff
@@ -68,7 +74,8 @@ public IndexerService(INetworkConfiguration networkConfiguration, HttpClient htt
         public async Task<List<ProjectIndexerData>> GetProjectsAsync(int? offset, int limit)
         {
             var indexer = _networkService.GetPrimaryIndexer();
-            var response = await _httpClient.GetAsync($"{indexer.Url}/api/query/Angor/projects?offset={offset}&limit={limit}");
+            var response =
```
dangershony, line 77: what has changed in this line?

src/Angor/Shared/Services/IndexerService.cs
```diff
@@ -225,5 +237,50 @@ public async Task<string> GetTransactionHexByIdAsync(string transactionId)
 
             return info;
         }
+
+        public async Task<string> GetGenesisBlockHashAsync(string indexerUrl)
+        {
+            try
+            {
+                var response = await _httpClient.GetAsync(indexerUrl);
```
dangershony, line 245: Shouldn't you be adding here the path? 
`api/query/block/index/0`

src/Angor/Shared/Services/IndexerService.cs
```diff
@@ -225,5 +237,50 @@ public async Task<string> GetTransactionHexByIdAsync(string transactionId)
...
+        }
+
+        public bool ValidateGenesisBlockHash(string fetchedHash, string expectedHash)
+        {
+            // Compare the first 64 characters (32 bytes in hex) of the fetched hash to the expected hash
+            return fetchedHash.StartsWith(expectedHash, StringComparison.OrdinalIgnoreCase);
+        }
+
+
+        public async Task<bool> CheckIndexerStatusAsync(string indexerUrl)
```
dangershony, line 267: I think we already have a way to check the indexer status look in the settings page for how we do that.

src/Angor/Shared/Services/IndexerService.cs
itailiors: nothing, just pressed enter there haha

src/Angor/Client/Pages/Settings.razor
```diff
@@ -471,39 +472,81 @@
...
+                settingsInfo.Indexers.Remove(addedIndexer);
+                _clientStorage.SetSettingsInfo(settingsInfo);
+                return;
+            }
+
+            // Validate genesis block hash if the indexer is online
+            var fetchedHash = await _indexerService.GetGenesisBlockHashAsync(newIndexerLink);
+            var expectedHash = _networkConfiguration.GetGenesisBlockHash();
+
+            if (string.IsNullOrEmpty(fetchedHash) || !_indexerService.ValidateGenesisBlockHash(fetchedHash, expectedHash))
```
dangershony, line 531: just do this check `string.IsNullOrEmpty` inside the method `ValidateGenesisBlockHash`

src/Angor/Shared/Services/IndexerService.cs
```diff
@@ -141,7 +151,8 @@ public async Task<string> PublishTransactionAsync(string trxHex)
             return response.ReasonPhrase + content;
         }
 
-        public async Task<AddressBalance[]> GetAdressBalancesAsync(List<AddressInfo> data, bool includeUnconfirmed = false)
+        public async Task<AddressBalance[]> GetAdressBalancesAsync(List<AddressInfo> data,
```
dangershony, line 154: try to must fix your ide to not add all this clutter

src/Angor/Shared/Services/IndexerService.cs
```diff
@@ -225,5 +239,40 @@ public async Task<string> GetTransactionHexByIdAsync(string transactionId)
 
             return info;
         }
+
+        public async Task<string> GetGenesisBlockHashAsync(string indexerUrl)
```
dangershony, line 243: I suggest that this method will also check if the indexer is online and also return the genesis hash.
then you can avoid the hack of adding to settings and then removing.
The code to check inline is simple it is just 
` var response = await _httpClient.GetAsync($"{uri}api/stats/heartbeat");`

src/Angor/Client/Pages/Settings.razor
```diff
@@ -471,39 +472,81 @@
...
             if (settingsInfo.Indexers.Any(a => a.Url == newIndexerLink))
             {
-                notificationComponent.ShowErrorMessage($"url exists {newIndexerLink}");
+                notificationComponent.ShowErrorMessage($"URL already exists: {newIndexerLink}");
                 return;
             }
 
-            settingsInfo.Indexers.Add(new SettingsUrl { Url = newIndexerLink, IsPrimary = !settingsInfo.Indexers.Any() });
+            // Add the new indexer temporarily
+            settingsInfo.Indexers.Add(new SettingsUrl
```
dangershony, line 503: All this code can go away and we should only call `inderxer.CheckIndexerNetwork` that will check online and also fetch block 0

itailiors:
@dangershony for each different indexer it seems like the api route is different, for example mempool doesnt have "query" like angor indexer.
how do we want to address that?
possible solution is to have route normalization or api mapping.
also we can just not support it (even though it might be on the same network)

links for genesis block hashes -
angor testnet indexer link : https://tbtc.indexer.angor.io/api/query/block/index/0
mempool testnet indexer link: https://mempool.space/testnet/api/block-height/0
mempool signet : https://mempool.space/signet/api/block-height/0

dangershony:
We still don't support mempool api

dangershony (approved):
Very good

PR #184, opened 2024-12-01
add cap on invest

itailiors:


itailiors:
because indexer is down i couldnt check the change, but its quite straight forward

Issue #172, opened 2024-11-13
Unlinkable Angor project identifiers 


dangershony:
To break the ability to link a transaction to the Angor protocol we should use the same techniques as silent payments.

dangershony:
https://github.com/MetacoSA/NBitcoin/issues/1240

dangershony:
This is the plan.
(inspired by the silent payment design [bip352](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki))

Creating the project identifier, the founder uses their private key to create a shared secret 
`A = founder key`
`B = angor key`
`P = project identifier key`

```
Let P = B + hash(a·B)·G
Encode P as an Angor identifier with bech32 prefix angor+pubkey (on the blockchain it will be a P2WPK output)
``` 

Spending the fee in the Angor Identifier address
```
Let p = b + hash(A·b)·G
```

p is the private key that can spend the coins in the Angor address
`P` can only generated by the founder and angor key.
`p` can only be generated by angor key.

PR #188, opened 2024-12-03
Unlinkable Angor project identifiers

dangershony:
> some of the code is borrowed from wasabi but it will get deleted

Solves this issue https://github.com/block-core/angor/issues/172

The class `ProjectIdentifierDerivation` is used to derive a shared secret between two parties.
Using the shared secret, the parties can derive a new public key that is unique to the two parties but only the receiver can derive the private key.

This PR will change the way we generate the Angor project identifier in such a way that it will be impossible to link an transaction that is an investment to the Angor key.

Thanks to @vernorvinge for suggesting this design to break the link between the founder and Angor addresses


dangershony:
The next thing to implement in the PR is to generate the several project identifiers using the founder keys when creating a wallet

dangershony:
This is not used yet and maybe never will but I will push this code so to have it ready in case we decide to go ahead and break the link between angor mainkey and projects

PR #190, opened 2024-12-04
Psbt test implantation for WalletOperations 

itailiors:


src/Angor/Shared/Utilities/BlockcoreNBitcoinConverter.cs
```diff
@@ -0,0 +1,40 @@
+using Blockcore.Consensus.TransactionInfo;
```
itailiors, line 1: should this be in utilities?

dangershony:
This pr is interesting but it is lower in priority right now

dangershony:
psbt is being implemented in other prs this can close

PR #191, opened 2024-12-05
Update the workflow with new flow

sondreb:
This reduces the number of workflows to 3: The existing Pull Request is kept; new Build and Release is added.

Build runs on commits to main. It will perform a publish (which automatically does restore and build), then run unit tests. This is run for Linux, Mac and Windows.

Then it will package up the results and upload as build artifacts. This makes them available for download in the individual build log. Then it installs the Rust tooling and performs a build for desktop apps. It does it from the binaries already built one single time. These are also uploaded as artifacts.

Then the process spawns into two: Publish (Release) and Deploy (Test). The Publish (Release) will create a new draft release, or update the existing, with the files from the build artifacts. The Deploy (Test) will download the Linux-artifact and publish to the "angor-test" repository "main" branch, which then gets deployed to GitHub Pages automatically.

When a release draft is published, the Release workflow will trigger. This will download the Linux package from release artifacts and publish to the gh-pages branch on the main Angor repo. It does not perform a new compilation; it uses the existing binaries from previous build.

The version on packages is 3 digits and it also adds the GitHub Workflow Run number as revision to the binaries built in the process. This means the actual build run is visible in the deployed/installed version of Angor.

Here is a screenshot that shows Angor (Test) with release 0.0.72.50 (run number 50), and Angor (Production) with release 0.0.71.48 and the Angor (Desktop) with release 0.0.71.48.

image


dangershony (approved):

dangershony:
@DavidGershony will also review and the push

DavidGershony (approved):

PR #193, opened 2024-12-05
Disable the minify steps in workflow

sondreb:
- Creates issues with security validation, because builds generates hashes for files and uses that to ensure correct files are loaded.


sondreb:
Should fix this issue: 
image

dangershony (approved):

PR #196, opened 2024-12-05
Dont try to decrypt a failed trx sig request the second time

dangershony:


src/Angor/Client/Pages/Signatures.razor
```diff
@@ -286,6 +286,16 @@
 
             foreach (var pendingSignature in signaturesRequests.Where(_ => _.AmountToInvest == null))
             {
+                if (string.IsNullOrEmpty(pendingSignature.TransactionHex))
```
DavidGershony, line 289: As we discussed rename Transaction Hex to Encrypted message and filter out nulls in the loop
Also we should perhaps add the decryption to the try catch so the whole flow doesn't fail

Issue #139, opened 2024-08-22
Project Key Overwritten When Creating a Project in Different Browsers with the Same Wallet


itailiors:
When a user creates a project in one browser (Browser A) and then goes to a different browser (Browser B) and recovers the wallet using the wallet words, fetches the projects, and creates a new project, an issue occurs. If the user returns to Browser A and attempts to create another project, the new project creation overrides the previously created project key instead of generating a new one. This leads to confusion and potential data loss as the system continues to build on the overridden project key.

Steps to Reproduce:

1. In Browser A, create a new project.
2. Switch to Browser B and recover the wallet using the wallet words.
3. Fetch the projects and create a new project in Browser B.
4. Return to Browser A and attempt to create another project.

Proposed Fixes:

- Refetch Projects on Create Button: Add logic to refetch the projects before initiating the creation of a new project to ensure the latest project data is available. (the logic is already added in another PR I opened #136 for issue #103 )
- Double-Check Before Posting Data: Refactor the project creation process so that no data is posted until all necessary data is collected and verified. This includes a double-check to ensure that a project with the same key has not already been created.

itailiors:
should be partially solved in #175

dangershony:
@DavidGershony still needs to verify the change in scanning

PR #197, opened 2024-12-06
fix for issues 166 and 139

itailiors:
to be tested

### DIFF
--- a/src/Angor/Client/Pages/Create.razor
+++ b/src/Angor/Client/Pages/Create.razor
@@ -19,6 +19,8 @@
 @inject IRelayService _RelayService
 @inject ILogger<Create> _Logger;
 @inject IHtmlStripperService HtmlStripperService;
+@inject IFounderKeyService FounderKeyService
+
 
 @inject IFounderTransactionActions _founderTransactionActions
 <NotificationComponent @ref="notificationComponent"/>
@@ -847,6 +849,21 @@
 
         try
         {
+            // Check if the FounderKey is in use and get the total keys used
+            var keyCheckResult = await FounderKeyService.CheckFounderKeyAsync(project.ProjectInfo.FounderKey);
+
+            if (keyCheckResult.IsKeyInUse)
+            {
+                notificationComponent.ShowErrorMessage("The FounderKey is already in use. Please use a different key.");
+                return; 
+            }
+
+            if (keyCheckResult.TotalKeysInUse >= 14)
+            {
+                notificationComponent.ShowErrorMessage("You cannot create more than 15 projects.");
+                return; 
+            }
+            
             var words = await passwordComponent.GetWalletAsync();
             var accountInfo = storage.GetAccountInfo(network.Name);
 
@@ -863,7 +880,6 @@
         }
         catch (Exception e)
         {
-            _Logger.LogError(e, e.Message);
             notificationComponent.ShowErrorMessage(e.Message,e);
         }
         finally
--- a/src/Angor/Client/Pages/Founder.razor
+++ b/src/Angor/Client/Pages/Founder.razor
@@ -34,14 +34,19 @@
 
                 </div>
             </div>
-            <button class="btn btn-border" @onclick="NavigateToCreateProject" disabled="@scanningForProjects">
-                <i>
-                    <Icon IconName="add"></Icon>
-                </i>
-                <span class="nav-link-text ms-1">
-                    @(scanningForProjects ? "Scanning..." : "Create Project")
-                </span>
-            </button>
+            <div class="tooltip-container" data-bs-toggle="tooltip" title="@GetCreateButtonTooltip()">
+                <button
+                    class="btn btn-border"
+                    @onclick="NavigateToCreateProject"
+                    disabled="@scanningForProjects || @(founderProjects.Count >= 14)">
+                    <i>
+                        <Icon IconName="add"></Icon>
+                    </i>
+                    <span class="nav-link-text ms-1">
+                        @(scanningForProjects ? "Scanning..." : "Create Project")
+                    </span>
+                </button>
+            </div>
 
         </div>
         <p class="mb-0 font-weight-normal text-sm mt-4">
@@ -214,5 +219,14 @@
 
         NavigationManager.NavigateTo("/create");
     }
+    private string GetCreateButtonTooltip()
+    {
+        if (founderProjects.Count >= 15)
+            return "You have reached the maximum number of projects. Please manage your existing projects.";
+        if (scanningForProjects)
+            return "Scanning in progress...";
+        return "Create a new project.";
+    }
+
 
 }
--- a/src/Angor/Client/Services/FounderKeyService.cs
+++ b/src/Angor/Client/Services/FounderKeyService.cs
@@ -1,0 +1,80 @@
+using Angor.Shared.Models;       
+using Angor.Shared.Services;     
+using Nostr.Client.Messages;     
+using Angor.Client.Storage;
+
+public class FounderKeyService : IFounderKeyService
+{
+    private readonly IWalletStorage _walletStorage;
+    private readonly IRelayService _relayService;
+    private readonly ISerializer _serializer;
+    
+
+
+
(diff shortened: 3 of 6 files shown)


src/Angor/Client/Services/FounderKeyService.cs
```diff
@@ -0,0 +1,80 @@
...
+        var inUseKeys = new HashSet<string>();
+
+        await _relayService.RequestProjectCreateEventsByPubKeyAsync(
+            keys.Keys.Select(k => k.NostrPubKey).ToArray(),
+            eventMessage =>
+            {
+                if (eventMessage.Kind == NostrKind.ApplicationSpecificData)
+                {
+                    var projectInfo = _serializer.Deserialize<ProjectInfo>(eventMessage.Content);
+                    inUseKeys.Add(projectInfo.FounderKey);
```
dangershony, line 39: this will fail if you get more then one result (multiple relays?)

dangershony: wait unless HashSet allows duplicates? but I think it does not or am I wrong?

src/Angor/Client/Pages/Create.razor
```diff
@@ -847,6 +849,21 @@
 
         try
         {
+            // Check if the FounderKey is in use and get the total keys used
+            var keyCheckResult = await FounderKeyService.CheckFounderKeyAsync(project.ProjectInfo.FounderKey);
```
dangershony, line 853: Putting this in the `CreatProjectTransaction` is probably not the right thing to do, we must know much before this point if the key was already used

dangershony (changes requested):
this must be review by @DavidGershony for the Nostr calls.
But also I am sure we do the check in the right place

src/Angor/Client/Pages/Founder.razor
```diff
@@ -34,14 +34,19 @@
...
             </div>
-            <button class="btn btn-border" @onclick="NavigateToCreateProject" disabled="@scanningForProjects">
-                <i>
-                    <Icon IconName="add"></Icon>
-                </i>
-                <span class="nav-link-text ms-1">
-                    @(scanningForProjects ? "Scanning..." : "Create Project")
-                </span>
-            </button>
+            <div class="tooltip-container" data-bs-toggle="tooltip" title="@GetCreateButtonTooltip()">
```
dangershony, line 37: can you make a separate pr just for issue 166?

src/Angor/Client/Services/FounderKeyService.cs
itailiors: When you add an item to a HashSet that already contains that item, it ignores it.
So it doesn't look like a problem if both relays return the same value. 

( ran this thing to check,

```
using System;
using System.Collections.Generic;

class Program
{
    static void Main()
    {
        HashSet<string> hashSet = new HashSet<string>();

        hashSet.Add("key1");
        hashSet.Add("key2");
        hashSet.Add("key1"); //duplicate
        hashSet.Add("key3");
        hashSet.Add("key2"); //dup

        foreach (var key in hashSet)
        {
            Console.WriteLine(key);
        }
    }
}
```
)

src/Angor/Client/Pages/Create.razor
itailiors: We do rescan before creating,
But there's a scenario this scenario.
1. Logged on browser A, B.
2. start create browser A.
3. Start create browser B.
4. Create browser B.
5. Create Browser A.

unless you put a rescan when creating, you won't know that there's another project created on the same key.
but also now that i think of it we will need to refresh the data in the create a project. (almost everything except project info no?)

I know that this is an edge case that is not expected (especially when we suspect that per user there will be probably only 1 project), but still this should be addressed in some way because we cannot have any loophole of over riding a project.

src/Angor/Client/Pages/Founder.razor
itailiors: done, 
but there's also a check that is a corelated to 166 in createtransaction that checks how many keys are already existing, and if there are more than 15 then fail.

(because of an issue like listed above)

itailiors: i'll also add that its needed because disabling the button in founder.razor isn't good enough, 
because people can "hack" and change the URL.
so there needs to be check in the createtransaction (or a check on oninitiallized in create.razor that doesn't show the create module if you have more than 15.)

src/Angor/Client/Services/FounderKeyService.cs
```diff
@@ -0,0 +1,80 @@
...
+        ISerializer serializer)
+    {
+        _walletStorage = walletStorage;
+        _relayService = relayService;
+        _serializer = serializer;
+    }
+
+    public async Task<bool> IsFounderKeyInUseAsync(string founderKey)
+    {
+        var keys = _walletStorage.GetFounderKeys();
```
DavidGershony, line 29: I'm not sure the relay is the right place to check, we need to check with the indexer if the transaction exists.

src/Angor/Shared/Services/RelayService.cs
```diff
@@ -89,6 +89,58 @@ public void RequestProjectCreateEventsByPubKey(Action<NostrEvent> onResponseActi
...
+                });
+            }
+
+            nostrClient.Send(new NostrRequest(subscriptionKey, new NostrFilter
+            {
+                Authors = nPubs,
+                Kinds = new[] { NostrKind.ApplicationSpecificData, NostrKind.Metadata },
+            }));
+
+            await tcs.Task; // Wait for the end of stream
```
DavidGershony, line 141: Why are we awaiting the tcs and not just allowing the callback thread to perfome the action we need?

src/Angor/Client/Pages/Create.razor
DavidGershony: We need to have a meeting about trusting the relays for knowing if the project was created or only look at the indexer to see if a transaciton exists

src/Angor/Client/Pages/Create.razor
```diff
@@ -847,6 +849,21 @@
...
+
+            if (keyCheckResult.IsKeyInUse)
+            {
+                notificationComponent.ShowErrorMessage("The FounderKey is already in use. Please use a different key.");
```
DavidGershony, line 857: In this case we need to "fix" the data in the local storage because the user can't select a founder key to use, perhaps add a message to rescan or wipe the storage and recover the wallet again.

src/Angor/Shared/Services/RelayService.cs
itailiors: I did it because of this 
https://learn.microsoft.com/en-us/dotnet/standard/asynchronous-programming-patterns/implementing-the-task-based-asynchronous-pattern

why shouldnt we need `TaskCompletionSource ` to wait for end of steam?

either way copilot removed it 😉

src/Angor/Client/Pages/Create.razor
itailiors: Okay, then i'll wait with this PR until we do that,
We should align on how much we trust the relays, and on which instances we can trust it.

Also, according to that we should check if there are other places in angor that trust the relay instead of the indexer?

itailiors:
created by copilot, pretty neat. 

The pull request #197 titled "fix for issue 139" includes the following changes:

- **Commits:**
  1. [fix for issues 166 and 139](https://github.com/block-core/angor/commit/04a9487dc4cc30bdcab23436a061cccd34939aad)
  2. [revert change for Founder.razor](https://github.com/block-core/angor/commit/12437e8c79448bacf39babc59d912e68011e6a47)
  3. [half act on commit, tbd](https://github.com/block-core/angor/commit/15766fe9d7b4f9b1f0e0831dc15e75966e948344)

- **Summary:**
  This PR addresses issue #139, making necessary code changes to fix the identified problems. The modifications include 141 additions and 1 deletion across 5 files. The specific changes involve fixing certain functionalities and reverting a change related to the `Founder.razor` file. The exact details of the changes can be reviewed in the commits linked above.

PR #201, opened 2024-12-06
Fix a nostr subscription registration bug

dangershony:
Check the dictionary _eoseCalledOnSubscriptionClients before adding a subscription class

I am not sure this is the correct fix because it might mean that we are not disposing of a subscription correctly
note this error returns until I refresh the page

### DIFF
--- a/src/Angor/Shared/Services/NostrCommunicationFactory.cs
+++ b/src/Angor/Shared/Services/NostrCommunicationFactory.cs
@@ -114,13 +114,15 @@ public class NostrCommunicationFactory : IDisposable , INostrCommunicationFactor
         return response;
     }
     
-    public void MonitoringEoseReceivedOnSubscription(string subscription)
+    public bool MonitoringEoseReceivedOnSubscription(string subscription)
     {
         _logger.LogInformation($"Started monitoring subscription {subscription}");
         var addedToSubscriptionClients = _eoseCalledOnSubscriptionClients.TryAdd(subscription, new List<string>());
 
         if(!addedToSubscriptionClients)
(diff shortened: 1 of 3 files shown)


src/Angor/Shared/Services/NostrCommunicationFactory.cs
```diff
@@ -117,7 +117,7 @@ public bool EoseEventReceivedOnAllRelays(string subscription)
     public void MonitoringEoseReceivedOnSubscription(string subscription)
     {
         _logger.LogInformation($"Started monitoring subscription {subscription}");
-        _eoseCalledOnSubscriptionClients.Add(subscription, new List<string>());
+        _eoseCalledOnSubscriptionClients.TryAdd(subscription, new List<string>());
```
DavidGershony, line 120: Better to add a boolean returned from the method and check it in the caller so we can identify where the duplication is from

dangershony: So you mean a `return TryAdd(...` ?

dangershony: Instead I logged a warning is this good enough?
this was called from `RelaySubscriptionsHandling.TryAddEoseAction`

DavidGershony: Return the boolean and we can handle it from the caller

dangershony: Handle how? log it at the caller instead?

DavidGershony: Yes, that way we can tell where the duplication happens and why

PR #210, opened 2024-12-12
Add Grids

SuperJMN:
To Browse and Wallet


src/Angor/Avalonia/AngorApp/Sections/Browse/Details/IProjectDetailsViewModel.cs
```diff
@@ -6,4 +6,13 @@ public interface IProjectDetailsViewModel : IProject
     string ShortDescription { get; }
     object Icon { get; }
     object Picture { get; }
+    public IEnumerable<Stage> Stages { get; }
+}
+
+public class Stage
```
dangershony, line 12: we might already have such types in the shared project, just remember later when we add that project we can use those types (or have to do lots of mappings)

dangershony (approved):

PR #215, opened 2024-12-12
Fixed bug in the founder project scanning

DavidGershony:
We weren't updating the project index which caused the wrong derivation when using it


dangershony (approved):
I can see how this made index get out of sync

Issue #204, opened 2024-12-09
Prevent Bypassing Restrictions by Manipulating URLs


itailiors:
We've noticed a problem where certain restrictions in the app can be bypassed by manually changing the URL. For example:

- If investing is disabled because a project's start date has passed, users can still navigate directly to /invest and potentially access the page
- Similarly, for /create, even if it's restricted in the UI, users can bypass the restriction by typing the URL.

To be fixed.

itailiors:
i'll note that if you don't have a wallet, it just send you to /wallet.
so there it is dealt with alright.

PR #217, opened 2024-12-13
Issue 204 - Blocking pages from server side

itailiors:
This is linked to issue #204


src/Angor/Client/Pages/Create.razor
```diff
@@ -587,11 +597,21 @@
         }
 
         var projects = storage.GetFounderProjects() ?? new List<FounderProject>();
+        projectCount = projects?.Count ?? 0;
```
dangershony, line 600: you forgot to look at `CreationTransactionId` for this

src/Angor/Client/Pages/Create.razor
```diff
@@ -587,11 +597,21 @@
...
+
         var keys = _walletStorage.GetFounderKeys();
 
-        if (projects.Count(p => !string.IsNullOrEmpty(p.CreationTransactionId)) >= keys.Keys.Count)
```
dangershony, line 592: Why is this exiting code not enough what are you trying to fix in this PR?

src/Angor/Client/Pages/Invest.razor
```diff
@@ -374,6 +374,7 @@ else
     TransactionInfo? signedTransaction;
     Transaction unSignedTransaction;
     string InvestorPubKey;
+    private ProjectValidationService validationService = new ProjectValidationService();
```
dangershony, line 377: you wanna use injection this is not common practice remember we always inject services on startup

src/Angor/Client/Models/ProjectValidationService.cs
```diff
@@ -0,0 +1,15 @@
+namespace Angor.Client.Models;
+
+public class ProjectValidationService
+{
+    public bool CanInvest(Project project)
```
dangershony, line 5: instead of a service what we did up to now is create a method on the Project class with the same logic

src/Angor/Client/Pages/Create.razor
itailiors: if you go directly to the /create in the URL even if you have more than 15 projects it will allow it.
Same with /invest (if the time is over, there's still that hack).

Nevertheless, you are correct here, it's the same check (i'll remove the redundant check)

src/Angor/Client/Pages/Invest.razor
```diff
@@ -935,4 +942,15 @@ else
     {
         showTransactionJsonModal = isVisible;
     }
+    
+    public bool CanInvest(Project project)
+    {
+        if (project == null) return false;
+
+        var now = DateTime.UtcNow;
+        if (now < project.ProjectInfo.StartDate) return false; // Investing period hasn't started
```
dangershony, line 951: no this is wrong `StartDate` is the date the project starts so it means we can only invest before the `StartDate` and we do not need to look at the `ExpiryDate`

the condition should be 

`now < project.ProjectInfo.StartDate return true`

dangershony: also add this just in case
`if (project.ProjectInfo == null) return false;`

src/Angor/Client/Pages/View.razor
```diff
@@ -315,16 +315,9 @@ else
...
-                    {
-                        <text>Invest Now</text>
-                    }
-                    else
-                    {
-                        <text>The investing period is over</text>
-                    }
+                        disabled="@(project == null || !CanInvest(project))"
+                        title="@(CanInvest(project) ? null : "Investment is not allowed for this project.")">
+                    Invest Now
```
dangershony, line 320: no I prefer how it was before just put text there instead of a button it is nicer I think

src/Angor/Client/Pages/View.razor
```diff
@@ -748,5 +742,16 @@ else
         string sanitizedInput = HtmlStripperService.StripHtmlTags(input);
         return new MarkupString(sanitizedInput);
     }
+    
+    public bool CanInvest(Project project)
+    {
+        if (project == null) return false;
+
+        var now = DateTime.UtcNow;
+        if (now < project.ProjectInfo.StartDate) return false; // Investing period hasn't started
```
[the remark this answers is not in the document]
itailiors: The original thought was because both in /view and /invest we should have the same conditions for someone who can invest that we should have a function (created it as a service) that both places use that determines it.
And that way if we want to add a condition, we will just add it in one place.

Also, this way it's more readable to understand what the conditions are to invest, and not scramble around the code trying to find all the conditions (right now there are little, but maybe in the future there will be more?)

src/Angor/Client/Pages/View.razor
```diff
@@ -748,5 +742,18 @@ else
         string sanitizedInput = HtmlStripperService.StripHtmlTags(input);
         return new MarkupString(sanitizedInput);
     }
+    
+    public bool CanInvest(Project project)
```
dangershony, line 746: hmm this is duplicate code I dont like that where could we put this?

src/Angor/Client/Pages/Create.razor
dangershony: ah I see yes this is probably needed ok

src/Angor/Client/Pages/View.razor
```diff
@@ -306,24 +306,17 @@ else
...
                 <button class="btn btn-border mb-3" data-cy="INVEST_BUTTON"
                         @onclick="InvestInProject"
-                        disabled="@(timeLeft <= 0 ? "disabled" : null)"
-                        title="@(timeLeft <= 0 ? "The investing period is over" : null)">
-                    @if (timeLeft > 0)
+                        disabled="@(project == null || !CanInvest(project))">
+                    @if (project == null || !CanInvest(project))
                     {
-                        <text>Invest Now</text>
+                    <text>The investing period is over</text>
```
dangershony, line 315: notice the spaces

PR #218, opened 2024-12-13
Check blockchain before allowing invest

itailiors:


src/Angor/Client/Pages/Invest.razor
```diff
@@ -413,22 +415,58 @@ else
...
+        {
+            
+            var words = await passwordComponent.GetWalletAsync();
+            var investorPubKey = _DerivationOperations.DeriveInvestorKey(words, project.ProjectInfo.FounderKey);
+
+            var isAlreadyInvested = await IsInvestorAlreadyInvestedAsync(project.ProjectInfo.ProjectIdentifier, investorPubKey);
+            if (isAlreadyInvested)
+            {
+                notificationComponent.ShowErrorMessage("You are already invested in this project.");
+                NavigationManager.NavigateTo($"/view/{ProjectId}");
```
itailiors, line 447: instead of throwing back to the view page, maybe just change the ui to say "already invested"

src/Angor/Client/Pages/Invest.razor
```diff
@@ -413,22 +415,58 @@ else
...
+        if (project != null)
+        {
+            
+            var words = await passwordComponent.GetWalletAsync();
+            var investorPubKey = _DerivationOperations.DeriveInvestorKey(words, project.ProjectInfo.FounderKey);
+
+            var isAlreadyInvested = await IsInvestorAlreadyInvestedAsync(project.ProjectInfo.ProjectIdentifier, investorPubKey);
+            if (isAlreadyInvested)
+            {
+                notificationComponent.ShowErrorMessage("You are already invested in this project.");
```
itailiors, line 446: shownotificationmessage can be a better fit

dangershony:
We need to carefully look at this PR, I understand what you are trying to do though.
What I don't like is the mixed logic, we already have code to pull my investments in the portfolio page so it means we have duplicated logic, perhaps instead we should redirect a user there or have a general method to scan for projects I have invested in.
We can also have two types of such a search a quick search that only looks in the blockchain and a deep search (like we have now in the portfolio page) that will also pull metadata.

Also remember that we are trying to reduce the dependency on the custom indexer endpoints calling `_IndexerService.GetInvestmentsAsync` is a custom endpoint

PR #219, opened 2024-12-13
Add Random Space Images for Nostr Profile Defaults on Testnet

itailiors:
Set default space-themed images for missing testnet Nostr profile banners and profile pictures.

### DIFF
--- a/src/Angor/Client/Pages/Create.razor
+++ b/src/Angor/Client/Pages/Create.razor
@@ -640,6 +640,13 @@
         // check if the project was already published to nostr
         nostrMetadataCreated = project.NostrProfileCreated;
         nostrApplicationSpecificDataCreated = !string.IsNullOrEmpty(project.ProjectInfoEventId);
+        
+        // Select a preset
+        if(!project.ProjectInfo.Stages.Any())
+        {
+            isPresetMode = true;
+            OnStagePresetChange(new ChangeEventArgs { Value = "preset1" });
+        }
     }
 
 
--- a/src/Angor/Client/Pages/Create.razor
+++ b/src/Angor/Client/Pages/Create.razor
@@ -599,36 +599,11 @@
         var projectsKeys = _derivationOperations.GetProjectKey(keys, latestProject?.ProjectIndex + 1 ?? 1);
 
         project = projects.FirstOrDefault(p => p.ProjectInfo.ProjectIdentifier == projectsKeys.ProjectIdentifier)
-                  ?? new FounderProject
-                  {
-                      ProjectIndex = projectsKeys.Index,
-                      Metadata = new ProjectMetadata(),
-                      ProjectInfo = new ProjectInfo
-                      {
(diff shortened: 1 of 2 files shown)


src/Angor/Client/Pages/Create.razor
```diff
@@ -724,6 +724,9 @@
 
         try
         {
+            // Ensure default images are set for Testnet
+            SetDefaultImagesForTestnet(project.Metadata);
```
dangershony, line 728: may I suggest we do that on startup?
also can we also do that for the presets? so a user just goes next next next and invest

src/Angor/Client/Pages/Create.razor
```diff
@@ -1172,4 +1175,45 @@
...
+            if (string.IsNullOrEmpty(metadata.Picture))
+            {
+                metadata.Picture = spacePhotos[random.Next(spacePhotos.Length)];
+            }
+        }
+    }
+    
+    private readonly string[] spacePhotos = new[]
+    {
+        "https://images-assets.nasa.gov/image/hubble-observes-one-of-a-kind-star-nicknamed-nasty_17754652960_o/hubble-observes-one-of-a-kind-star-nicknamed-nasty_17754652960_o~thumb.jpg?as=webp",
```
DavidGershony, line 1199: Are we allowed to use these images freely?

itailiors: as far as i read from here, it seems like we can.

https://www.nasa.gov/nasa-brand-center/images-and-media/

src/Angor/Client/Pages/Create.razor
itailiors: When you say startup, you mean when you go into the create-project there will be these links?
And also yeah we can do it for presets, should I make it random or a pre-determined?

dangershony: yeah random put some images on the correct boxes when we land on the page if it is testnet a user can always just change it

src/Angor/Client/Pages/Create.razor
```diff
@@ -616,6 +616,21 @@
                       }
                   };
 
+        // Set default images for Testnet
+        var random = new Random();
```
dangershony, line 620: whats going on with the spacing here?

src/Angor/Client/Pages/Create.razor
```diff
@@ -625,6 +615,13 @@
         // check if the project was already published to nostr
         nostrMetadataCreated = project.NostrProfileCreated;
         nostrApplicationSpecificDataCreated = !string.IsNullOrEmpty(project.ProjectInfoEventId);
+        
+        // Select a preset
+        if(!project.ProjectInfo.Stages.Any())
```
dangershony, line 620: calling this so late means the presets dont get saved to local storage but it is ok

dangershony (approved):

PR #233, opened 2024-12-23
Add Wallet flow. Refine project Invest (stages).

SuperJMN:
## Wallet Section
- It's now possible to complete a Send operation (test model)
- Receive shows the receive address

## Invest Section (Browse project)
- Refinements (stage breakdown)

## Misc
- Model interfaces reside in their own project
- Removed Browser project for now
- Better componentalization


src/Angor/Avalonia/AngorApp/Sections/Wallet/BroadcastedTransactionDesign.cs
```diff
@@ -0,0 +1,17 @@
...
+using CSharpFunctionalExtensions;
+
+namespace AngorApp.Sections.Wallet;
+
+public class BroadcastedTransactionDesign : IBroadcastedTransaction
+{
+    public string Address { get; init; }
+    public decimal FeeRate { get; set; }
+    public decimal TotalFee { get; set; }
+    public decimal Amount { get; init; }
```
dangershony, line 12: One thing that we have discuss earlier is that perhaps we should always represent amounts as sats and never as Bitcoin (decimal) only when we display we make the conversion

dangershony, line 12: One thing that we have discuss earlier is that perhaps we should always represent amounts as sats and never as Bitcoin (decimal) only when we display we make the conversion

SuperJMN: Yes! It that that is quite reasonable. I agree with that. Sats is a coherent way to hold amounts. In the GUI, we could later convert it to our needs.

dangershony (approved):

PR #236, opened 2024-12-27
Notify founder when there is a new investment 

itailiors:
added logic like in view.razor to scan for new investments with ScanForNewInvestmentsAsync. showed investment stats (total investors and total raised) in the project cards. used local storage to keep and compare stats, so we can detect new investments and notify the founder. this keeps things consistent with how tracking works in view.razor.

image

### DIFF
--- a/src/Angor/Client/Components/FounderProjectItem.razor
+++ b/src/Angor/Client/Components/FounderProjectItem.razor
@@ -2,31 +2,26 @@
 @using Angor.Client.Models
 @using Angor.Client.Storage
 @using System.Text.RegularExpressions
- 
+@using Blockcore.NBitcoin
+
 
 @inject IRelayService RelayService;
 @inject IClientStorage Storage;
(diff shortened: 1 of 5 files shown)


itailiors:
I'll add that adding a new event and creating that nostr event when investing and listening on founder felt like unwanted nostr calls. 
But on the other side there are 2 disadvantages here,
1) this utilizes local storage (that is bound to be refactored)
2) if someone recovers and scans the project, he will have a bunch of notifications (i don't think this is that big of a deal)

@dangershony @DavidGershony what's your thought of this approach instead

itailiors:
Also, I'll add that it felt like a good enhancement to the page imo.
If you think that it's better to create a new nostr event it shouldn't be a problem because I'm working on that on a different branch to inform the investor when his UTXO is getting spent.

dangershony:
We already have that functionality based on new messages in nostr so I think this can be closed

PR #237, opened 2024-12-27
Add Create Wallet + Amount Entry Improvement

SuperJMN:
This makes it possible to navigate around the application and test the existing flows:

- Browse: Invest (WIP).
- Wallet: Send/Receive.
- Wallet: Create wallet.

Extra
- Improves BTC address validation.
- Model uses sats (ulong) for amounts and fees.

It also includes a new control to allow users enter amounts in either sats or BTC.


src/Angor/Avalonia/AngorApp.Model/BitcoinAddressValidator.cs
```diff
@@ -0,0 +1,269 @@
+namespace AngorApp.Model;
+
+using System.Numerics;
+using System.Security.Cryptography;
+
+public class BitcoinAddressValidator
```
dangershony, line 6: I don't mind bringing code here but NBitcoin has such a validator no? but it is ok

dangershony (approved):

src/Angor/Avalonia/AngorApp.Model/BitcoinAddressValidator.cs
SuperJMN: Ah, yes! I created this helper using AI just to add some quick validation. It think it would be good to switch to existing NBitcoin code later on, for example, as soon as we start using the Shared project. What do you think?

PR #238, opened 2025-01-04
add service for refreshing the utxo's and utilize it in Create.razor

itailiors:


itailiors:
is there anywhere else in the project we're supposed to do that?
in invest?

dangershony:
> is there anywhere else in the project we're supposed to do that?
in invest?

yes probably also in spend and maybe also recover

dangershony:
@itailiors this was done in another PR, so we can close this one.

dangershony:
hmm I am not sure f I called the new refresh method form the create can you double check that when you have a moment

itailiors:
> hmm I am not sure f I called the new refresh method form the create can you double check that when you have a moment
verified 👍

PR #239, opened 2025-01-07
Add New Wallet Wizard

SuperJMN:
This adds the actual wizard


sondreb (dismissed):
Reviewed the code without running it yet, thanks for this PR.

- One future comment is to avoid using spaces ("_") in source code. Instead of calling it step "Step_3", it should be "Step3".

src/Angor/Avalonia/AngorApp/Sections/Wallet/Create/Step_2/SeedWordsViewModel.cs
```diff
@@ -0,0 +1,36 @@
+using System.Linq;
+using System.Reactive.Linq;
+using AngorApp.Model;
+using AngorApp.Sections.Browse;
+using CSharpFunctionalExtensions;
+using ReactiveUI.SourceGenerators;
+using ReactiveUI.Validation.Helpers;
+
+namespace AngorApp.Sections.Wallet.Create.Step_2;
```
dangershony, line 9: As suggested we never really `_` for types or ns only maybe for params, rename this to Step2

SuperJMN:
> Reviewed the code without running it yet, thanks for this PR.
> 
> * One future comment is to avoid using spaces ("_") in source code. Instead of calling it step "Step_3", it should be "Step3".

Thanks! I've replaced conflicting namespaces already.

dangershony (approved):

PR #251, opened 2025-01-21
Add "Get Projects" and "Find By Id" from Production Service

SuperJMN:
Implement service to get projects from the real service and to find a given project by ID.

Includes the refactors and improvements that get this done.


src/Angor/Avalonia/Angor.Model.Implementation/DependencyFactory.cs
```diff
@@ -0,0 +1,37 @@
+using Angor.Shared.Services;
+using Microsoft.Extensions.Logging;
+using Nostr.Client.Client;
+
+namespace Angor.Model.Implementation;
+
+public static class DependencyFactory
+{
+    public static IRelayService GetRelayService(ILoggerFactory loggerFactory)
```
dangershony, line 9: just wondering why not use the standard dependency injection?

src/Angor/Avalonia/Angor.Model.Implementation/Projects/ProjectService.cs
```diff
@@ -0,0 +1,85 @@
...
+                observer.OnNext,
+                observer.OnCompleted,
+                projectIndexerDatas.Select(x => x.NostrEventId).ToArray()
+            );
+
+            return Disposable.Empty;
+        });
+    }
+
+    private IObservable<(string, ProjectMetadata)> ProjectMetadatas(IEnumerable<ProjectInfo> projectInfos)
```
dangershony, line 76: this looks like an interesting pattern that we should consider using everywhere

src/Angor/Avalonia/Angor.Model.Implementation/DependencyFactory.cs
SuperJMN: Good question! I chose this approach because the number of dependencies is still relatively small and adding a DI container felt like overengineering at this stage. However, if you think we should switch to standard DI, I can prepare a PR with the changes.

dangershony: I am not sure what DI you had in mind to use but the default on the web code is `Microsoft.Extensions.DependencyInjection.ServiceProvider`

image

So probably we just use the same one

src/Angor/Avalonia/Angor.Model.Implementation/NetworkConfiguration.cs
```diff
@@ -0,0 +1,231 @@
+using Angor.Shared;
+using Angor.Shared.Models;
+using Blockcore.Networks;
+
+namespace Angor.Model.Implementation;
+
+// TODO: This is copied from Angor.Client
+public class NetworkConfiguration : INetworkConfiguration
```
dangershony, line 8: hmm I dont like this, but it is temporary code then I guess it is ok for now

src/Angor/Avalonia/Angor.Model.Implementation/Projects/Stage.cs
```diff
@@ -0,0 +1,11 @@
+using AngorApp.Model;
+
+namespace Angor.Model.Implementation.Projects;
+
+public class Stage : IStage
+{
+    public DateOnly ReleaseDate { get; set; }
+    public uint Amount { get; set; }
```
dangershony, line 8: money we represent as long normally

dangershony: unless this is the stage representation so it is a percent in which case it should be a decimal

src/Angor/Avalonia/AngorApp/Sections/Browse/ProjectLookup/ProjectLookupViewModel.cs
```diff
@@ -0,0 +1,67 @@
...
+                return maybeProject.Map<IProject, IList<IProjectViewModel>>(project =>
+                {
+                    var vm = new ProjectViewModel(walletProvider, project, navigator, uiServices);
+                    return new List<IProjectViewModel> { vm };
+                });
+            }
+        );
+
+        lookupResultsHelper = Lookup
+            .ObserveOn(RxApp.MainThreadScheduler)
```
dangershony, line 46: why do you need to use a schedular `MainThreadScheduler`? is it because there are calls over the wire?

src/Angor/Avalonia/Angor.Model.Implementation/Projects/Stage.cs
SuperJMN: It's the stage representation and it's a number in sats internally, that can be converted to whatever we need in the UI (BTC or sats)

src/Angor/Avalonia/Angor.Model.Implementation/NetworkConfiguration.cs
SuperJMN: Yes, that's why I decided to copy it

src/Angor/Avalonia/Angor.Model.Implementation/DependencyFactory.cs
SuperJMN: I tried to use use of Microsoft.Extensions.DependencyInjection. It's done and it works perfectly, but I think that change should be done in a follow-up PR. It's a bit of scope and I'd like to keep it separated. What do you think?

src/Angor/Avalonia/Angor.Model.Implementation/Projects/Project.cs
```diff
@@ -0,0 +1,20 @@
+using AngorApp.Model;
+
+namespace Angor.Model.Implementation.Projects;
```
DavidGershony, line 3: This model belongs to the UI only right? perhaps the namespace should reflect that?

SuperJMN: Yes, absolutely. I've modified the namespaced to reflect that.

dangershony (approved):

PR #252, opened 2025-01-21
New UI

miladsoft:


dangershony (approved):
I did not test it, I will test it all when UI is done

PR #255, opened 2025-02-04
Memool-end-to-end

DavidGershony:
Fixed issues with mempool implementation after integration testing

### DIFF
--- a/src/Angor/Shared/Services/MempoolSpaceIndexerApi.cs
+++ b/src/Angor/Shared/Services/MempoolSpaceIndexerApi.cs
@@ -15,6 +15,8 @@ public class MempoolSpaceIndexerApi : IIndexerService
     private readonly HttpClient _httpClient;
     private readonly INetworkService _networkService;
 
+    private const string AngorApiRoute = "/api/V1/query/Angor";
+
     public MempoolSpaceIndexerApi(ILogger<MempoolSpaceIndexerApi> logger, INetworkConfiguration networkConfiguration, HttpClient httpClient, INetworkService networkService)
     {
         _logger = logger;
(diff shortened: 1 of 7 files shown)


src/Angor/Shared/Services/MempoolSpaceIndexerApi.cs
```diff
@@ -0,0 +1,436 @@
...
+                });
+            }
+        }
+        
+        return response.ToArray();
+    }
+
+    
+
+    //To use if the UTXO endpoint works on the mempool indexer
```
dangershony, line 246: Can delete this I guess?

src/Angor/Shared/Services/NetworkService.cs
```diff
@@ -100,7 +100,10 @@ public async Task CheckServices(bool force = false)
                     try
                     {
                         var uri = new Uri(indexerUrl.Url);
-                        var response = await _httpClient.GetAsync($"{uri}api/stats/heartbeat");
+                        
+                        var blockUrl = Path.Combine(uri.AbsoluteUri, "api", "v1", "block-height", "0");
```
dangershony, line 104: This is very different way to check status, it will just return genesis, or is it only to test the api is available?

src/Angor/Shared/WalletOperations.cs
```diff
@@ -539,7 +539,7 @@ public async Task<IEnumerable<FeeEstimation>> GetFeeEstimationAsync()
             var feeEstimations = await _indexerService.GetFeeEstimationAsync(blocks);
 
             if (feeEstimations == null || (!feeEstimations.Fees?.Any() ?? true))
-                return blocks.Select(_ => new FeeEstimation{Confirmations = _,FeeRate = 10000 / _});
+                return blocks.Select(_ => new FeeEstimation{Confirmations = _,FeeRate = 10000 / _}); // default to 1 satoshi per byte for 10 blocks and 10 satoshi for 1 block
```
dangershony, line 542: the only change here is the comment?

src/Angor/Shared/Services/MempoolSpaceIndexerApi.cs
```diff
@@ -107,12 +107,15 @@ private class MempoolTransaction
     private class Outspent
     {
         public bool Spent { get; set; }
+        public string Txid { get; set; }
```
DavidGershony, line 110: Why did you add these parameters if we don't use them?

PR #263, opened 2025-02-15
New UI

miladsoft:


src/Angor/Client/Pages/Create.razor
```diff
@@ -1181,19 +1168,30 @@
...
+            {
+                if (sortedStages[i].ReleaseDate >= sortedStages[i + 1].ReleaseDate)
+                {
+                    notificationComponent.ShowErrorMessage($"Stage {i + 2} must be after stage {i + 1}");
+                    return false;
+                }
+            }
+
+            // Keep stages sorted
+            project.ProjectInfo.Stages = sortedStages;
```
dangershony, line 1192: I don't like this so much as we are making change on behalf of the user, we should give a warning and let the user fix it.

src/Angor/Client/Pages/Signatures.razor
```diff
@@ -292,17 +285,54 @@ else
                     var sigResJson = await encryption.DecryptNostrContentAsync(
                         nostrPrivateKeyHex, pendingSignature.investorNostrPubKey, pendingSignature.EncryptedMessage);
 
-                    pendingSignature.SignRecoveryRequest = serializer.Deserialize<SignRecoveryRequest>(sigResJson);
-
-                    if (pendingSignature.SignRecoveryRequest is null)
+                    if (!string.IsNullOrEmpty(sigResJson))
```
dangershony, line 288: no no Milad this is way too complicated code, we should role back to the code that was committed in my pr, there is no need for regex checks here

dangershony: I understand now, you are trying to support backwards compatibility.
I would argue we are on testnet there is no reason to support backwards compatibility on test data.
Projects that cant deserialize we just treat as spam messages.

dangershony (approved):

Issue #266, opened 2025-02-19
Bug in stages - Project creation


itailiors:
currently, you can create a project with stages with 0% and negative percent.

![Image](https://github.com/user-attachments/assets/2cdff2c5-a9b7-4c44-86a7-3db6c8262a35)

itailiors:
https://test.angor.io/view/angor1q584c5x7rchgmn6n8ftas6kh8l0xv78n7qk5506

itailiors:
also you can create a stage with more than 100% with this bug (so that should also be added as defense code)

https://test.angor.io/view/angor1q5qy503v6mxdk58l8x7j7naq2d3fvuwrq3glk4s

![Image](https://github.com/user-attachments/assets/d93682f3-93d2-4143-8b13-62628154ce42)

PR #267, opened 2025-02-19
fixing issue 266

itailiors:
#266

### DIFF
--- a/src/Angor/Client/Pages/Create.razor
+++ b/src/Angor/Client/Pages/Create.razor
@@ -1121,6 +1121,15 @@
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


src/Angor/Client/Pages/Create.razor
```diff
@@ -1121,6 +1121,15 @@
             return false;
         }
 
+        foreach (var stage in project.ProjectInfo.Stages)
```
dangershony, line 1124: I was sure we had this check

dangershony: ah yeah see above `The stages must sum to 100%` is it no the result?

dangershony (approved):

PR #268, opened 2025-02-19
UI improvements

miladsoft:


dangershony (approved):
Good I did not test it, we will test it as part of the full testing plan, but it is mostly html changes

PR #274, opened 2025-02-27
Implement WalletAppService

SuperJMN:
Adds a working implementation of WalletAppService.

It's still not used in the GUI, but will be used in a follow-up PR.


src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/AesWalletEncryption.cs
```diff
@@ -0,0 +1,90 @@
...
+using CSharpFunctionalExtensions;
+
+namespace Angor.Wallet.Infrastructure.Impl;
+
+public class AesWalletEncryption : IWalletEncryption
+{
+    private const int Iterations = 100000;
+    private const int KeySize = 256;
+
+    public async Task<Result<WalletData>> Decrypt(EncryptedWallet encryptedWallet, string encryptionKey)
```
DavidGershony, line 14: I prefer to have the word async in the end of an async method, perhaps we should bring this up with the others and have a consensus.

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/AesWalletEncryption.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        byte[] encryptedData;
+        using (var aes = Aes.Create())
+        {
+            aes.Key = key;
+            aes.IV = iv;
+
+            using var msEncrypt = new MemoryStream();
+            using (var csEncrypt = new CryptoStream(msEncrypt, aes.CreateEncryptor(), CryptoStreamMode.Write))
+            using (var writer = new StreamWriter(csEncrypt, Encoding.UTF8))
```
DavidGershony, line 65: you can serialize directly to the crypto stream
await JsonSerializer.SerializeAsync(csEncrypt, walletData);

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/NetworkExtensions.cs
```diff
@@ -0,0 +1,38 @@
+using Angor.Wallet.Domain;
+using NBitcoin;
+
+namespace Angor.Wallet.Infrastructure.Impl;
+
+public static class NetworkExtensions
+{
+    public static Network ToNBitcoin(this BitcoinNetwork bitcoinNetwork)
```
DavidGershony, line 8: Not really crazy about exposing NBitcoin types like this, but if we have to...

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/WalletAppService.cs
```diff
@@ -0,0 +1,266 @@
...
+            var walletWords = new WalletWords { Words = seed, Passphrase = passphrase };
+            var transactions = new List<BroadcastedTransaction>();
+                
+            var accountInfo = _walletOperations.BuildAccountInfoForWalletWords(walletWords);
+                
+            // Update existing UTXOs information
+            await _walletOperations.UpdateDataForExistingAddressesAsync(accountInfo);
+            await _walletOperations.UpdateAccountInfoWithNewAddressesAsync(accountInfo);
+
+            foreach (var address in accountInfo.AllAddresses())
```
DavidGershony, line 63: This can be optimised later on but looks correct

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/WalletAppService.cs
```diff
@@ -0,0 +1,266 @@
...
+            return Result.Failure<Address>($"Error getting receive address: {ex.Message}");
+        }
+    }
+
+    public async Task<Result<TxId>> SendAmount(WalletId walletId, Amount amount, Address address, DomainFeeRate feeRate)
+    {
+        if (walletId != SingleWalletId)
+            return Result.Failure<TxId>("Invalid wallet ID");
+        
+        var finalFeeRate = feeRate.SatsPerVByte * 10000;
```
DavidGershony, line 185: Why is this multiplied? if I understand correctly it should be * 1000 for sats per kilobyte

SuperJMN: I tried to copy the behavior of the UI, but I can completely wrong.

Given this:

image

The sendInfo is this:

image

I'm not sure about the conversion, to be honest. What's the correct way to calculate?

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/NetworkExtensions.cs
SuperJMN: I understand your concern about exposing **NBitcoin** types, but this implementation is specifically designed for the infrastructure layer where we need this conversion to interact with the library.
The `ToNBitcoin` method is necessary to translate between our domain model (`BitcoinNetwork`) and NBitcoin's `Network` type, but it's encapsulated in this class and not exposed outside the infrastructure layer. 
There's no way for these conversions to reach the upper layers of the application, which maintains our domain isolation while allowing us to work with the **NBitcoin** library where needed.

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/WalletAppService.cs
SuperJMN: We can implement any caching mechanism to handle those requests later :)

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/AesWalletEncryption.cs
SuperJMN: You're so right! Thanks for the correction! :D

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/AesWalletEncryption.cs
SuperJMN: I see your point about the Async suffix. Yes, it's in Microsoft's guidelines, but I'm not a fan of it for practical reasons - it creates redundancy and unnecessarily long names, especially in interfaces where almost every method returns a Task.
That said, if we want to standardize this across our codebase, I'm fine changing it. Should we bring this up with the team to decide on a consistent approach?

SuperJMN: Several popular .NET projects (like MediatR, Octokit, NServiceBus) don't follow this convention either. But I'm happy to go with whatever standard we decide on as a team.

src/Angor/Avalonia/Angor.Wallet/Infrastructure/WalletServices.cs
```diff
@@ -0,0 +1,39 @@
...
+public static class WalletServices
+{
+    public static ServiceCollection Register(ServiceCollection services, ILogger logger)
+    {
+        var loggerFactory = LoggerFactory.Create(builder => builder.AddConsole());
+        
+        services.TryAddSingleton<IStore>(new InMemoryStore());
+        logger.Warning("Registering default Store: {Store}", services.First(x => x.ServiceType.IsAssignableTo(typeof(IStore))).ServiceType);
+        
+        services.AddSingleton<IWalletAppService, WalletAppService>();
```
DavidGershony, line 23: Why do you register everything as singleton?

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/AesWalletEncryption.cs
DavidGershony: Keep it like this but let's bring it up in the next team meeting so we are all on the same page - consistency is more important

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/WalletAppService.cs
DavidGershony: I think the sendinfo in this case has a high fee to get it in the next block, but if it is sats ber byte and we convert it to sats per kilobyte it should be * 1000

src/Angor/Avalonia/Angor.Wallet/Infrastructure/WalletServices.cs
SuperJMN: Good question! I've double reviewed registrations and improved it a bit, and made it more readable :)

Well, I reviewed all the registrations in the Avalonia application and I believe the singleton approach is appropriate for our desktop application architecture because:

1. Most of our services are actually stateless or maintain immutable state once initialized. For example, NetworkService, IndexerService, WalletSecurityContext, and AesWalletEncryption all perform operations based on input parameters without maintaining changing state between calls. I'm not 100% sure in case of IndexerService, though. If it's stateful, we would need to think how that could affect the application.
2. Our application is single-user, desktop-based, so we don't have the thread safety concerns that would exist in a multi-user web environment.
3. I've used transient registration for ViewModels and UI components that do maintain state specific to user interactions or UI elements.

I'd be happy to review specific services if you have concerns about particular implementations, but I believe that for most of our infrastructure and application services, singleton registration is the appropriate choice given their stateless design.

Have you found any specific issues? I admit I tend to use singletons by default and adjust when problems appear, so your question definitely helped me review our approach more carefully.

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/WalletAppService.cs
SuperJMN: Fixed!

src/Angor/Avalonia/Angor.Wallet/Infrastructure/WalletServices.cs
DavidGershony: I will approve it, but lets talk about it because I don't see the benefit of forcing  the services lifetime to be that long, and memory should not be an issue as well.

DavidGershony (approved):

Issue #269, opened 2025-02-20
Fix the derivation


dangershony:
OLD
**Angor Key** - `A well known xpubkey`
**Founder key**   - `m / 5' / n'`   
**upi - Unique Project Identifier** - `Low64(hash256(founderPubKey))`  
**Project Identifier** - `Bech32(ExtPubkey(AngorKey).Derive(upi))`  
**Founder Recovery Key**   - `m / 6' / n'`  
**Founder/Investor Nostr Key (NIP06)** - `"m / 44' / 1237' / n / 0 / 0`  
**Investor Key**   - `m / 5' / 0' / upi' / 1'`  
**Investor Lead Secret**  - `m / 5' / 0' / upi' / 4'`  

n = an index of the project number created by the founder  
upi = an integer extracted from the lower part of the sha256 of the founder key

It needs to be

**Angor Key**                                               - `A well known xpubkey`
**Founder key**                                            - `m / 5' / n'`   
**upi - Unique Project Identifier**                - `Low64(hash256(founderPubKey))`  
**Project Identifier**                                     - `Bech32(ExtPubkey(AngorKey).Derive(upi))`  
**Founder/Investor Nostr Key (NIP06)**      - `m / 44' / 1237' / upi' / 0 / 0`  
**Founder Recovery Key**                            - `m / 5' / 0' / upi' / 1'`
**Investor Lead Secret**                               - `m / 5' / 0' / upi' / 2'`  
**Investor Key**                                            - `m / 5' / 0' / upi' / 3'`

PR #276, opened 2025-03-03
Fix the derivation path on all keys

dangershony:
This PR brings the code on how we derive keys to be as expected in the specification
https://github.com/block-core/angor/issues/269

Once this PR is pushed the BCIP will be updated.
https://github.com/block-core/bcips/blob/main/bcip-0005.md#key-derivation-path

### DIFF
--- a/src/Angor.Test/ProtocolNew/FounderTransactionActionTest.cs
+++ b/src/Angor.Test/ProtocolNew/FounderTransactionActionTest.cs
@@ -97,11 +97,10 @@ public class FounderTransactionActionTest : AngorTestData
     [Fact]
     public void SignInvestorRecoveryTransactions_CreatesValidSignatures()
     {
-        var words = new WalletWords
-            { Words = "sorry poet adapt sister barely loud praise spray option oxygen hero surround" };
+        var words = new WalletWords { Words = "sorry poet adapt sister barely loud praise spray option oxygen hero surround" };
         var projectInvestmentInfo = GivenValidProjectInvestmentInfo(words);
 
-        var founderRecoveryPrivateKey = _derivationOperations.DeriveFounderRecoveryPrivateKey(words, 1);
+        var founderRecoveryPrivateKey = _derivationOperations.DeriveFounderRecoveryPrivateKey(words, projectInvestmentInfo.FounderKey);
 
         var investmentTrxHex =
             "010000080005c0c62d0000000000160014e503a24793c82bf7f7eb18cfca6589df1360dcf40000000000000000236a21038a7eedf38d874799c0d7579d5f08d605ca039da7f6dc7c57e6abd82f0f380334e0930400000000002251207aade2c416ca565c1b66041eac56d707a26aa8bab7d217190ecf3aa57899c9a360e316000000000022512048dc0a52f43379c6515962a40eec91d183582c8e2c861d612e4e564617d08a67804f120000000000225120f57ad9ed9e0fb880cda0847cfea668a4c3ccb1b9a08bdbcc3f029e8ad0380c9d00000000";
--- a/src/Angor/Client/Pages/Invest.razor
+++ b/src/Angor/Client/Pages/Invest.razor
@@ -630,7 +630,7 @@ else
             {
                 var words = await passwordComponent.GetWalletAsync();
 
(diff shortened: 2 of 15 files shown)


src/Angor/Client/Pages/Invest.razor
```diff
@@ -1032,10 +1032,11 @@ else
     private async Task SaveInvestmentsListToNostrAsync()
     {
         var words = await passwordComponent.GetWalletAsync();
-        var rootNostrPrivateKey = await _derivationOperations.DeriveProjectNostrPrivateKeyAsync(words, 0);
-        var rootNostrPrivateKeyHex = Encoders.Hex.EncodeData(rootNostrPrivateKey.ToBytes());
+        var storageAccountKey = _derivationOperations.DeriveNostrStoragePubKeyHex(words);
```
DavidGershony, line 1035: It would help to add the PubKey in the name so it'a clear

src/Angor/Client/Pages/Unfunded.razor
```diff
@@ -613,12 +613,12 @@
     {
         try
         {
-            var key = DerivationOperations.DeriveFounderRecoveryPrivateKey(words, FounderProject.ProjectIndex);
+            var key = DerivationOperations.DeriveFounderRecoveryPrivateKey(words, FounderProject.ProjectInfo.FounderKey);
```
DavidGershony, line 616: perhaps we should keep the UPI as a property on the project info?

src/Angor/Client/Pages/Invest.razor
dangershony: Yeah but I hid this complexity in the method on purpose, better no?

src/Angor/Shared/Services/MempoolSpaceIndexerApi.cs
```diff
@@ -115,7 +115,7 @@ private class Outspent
     public async Task<List<ProjectIndexerData>> GetProjectsAsync(int? offset, int limit)
     {
         var indexer = _networkService.GetPrimaryIndexer();
-        var response = await _httpClient.GetAsync($"{indexer.Url}{AngorApiRoute}/projects?offset={offset ?? 0}&limit={limit}");
+        var response = await _httpClient.GetAsync($"{indexer.Url}{AngorApiRoute}/projects?offset={offset ?? null}&limit={limit}");
```
DavidGershony, line 118: Null breaks this in the mempool side, also we don't need to default to null if it is null

DavidGershony (dismissed):
Looks really good, with minor comments.

src/Angor/Client/Pages/Unfunded.razor
dangershony: Nah it just complicates things and that parameter is only ever used for derivation

src/Angor/Shared/DerivationOperations.cs
```diff
@@ -251,14 +240,46 @@ public uint DeriveProjectId(string founderKey)
...
+        ExtKey extendedKey = GetExtendedKey(walletWords);
+
+        var path = $"m/44'/1237'/1'/0/0";
+
+        ExtKey extKey = extendedKey.Derive(new KeyPath(path));
+
+        return extKey.PrivateKey;
+    }
+
+    public string DeriveNostrStoragePassword(WalletWords walletWords)
```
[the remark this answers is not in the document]
dangershony: Perhaps the name should be different?
`DeriveNostrStorageAccountPassword`

src/Angor.Test/ProtocolNew/FounderTransactionActionTest.cs
```diff
@@ -200,30 +197,32 @@ public void SpendFounderStage_TestUsingInvestmentOperations(int stageNumber)
             transactionList.Select(_ => _.Outputs.AsIndexedOutputs().ElementAt(stageNumber + 1).ToCoin()));
     }
 
+    //[Fact]
```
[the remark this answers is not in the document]
dangershony: This is not a test, it is a method to generate test data, I will put a comment

src/Angor/Shared/Services/MempoolSpaceIndexerApi.cs
dangershony: this is a problem then, how do we get the latest on the mempool indexer?

dangershony: fixed

PR #277, opened 2025-03-04
Add DynamicWallet and complete Wallet Features

SuperJMN:
- This PR adds **DynamicWallet** as implementation of IWallet and completes the wallet features. Some views have been modified and some adjustments were needed to complete the Wallet Features.

- This pull request is a follow-up from this https://github.com/block-core/angor/pull/274. 

We can address any issues or refinements in other PRs to make it easier to review.


DavidGershony (changes requested):
Can you rebase unto main so it's easier to see the changes from the other PR?

src/Angor/Avalonia/Angor.UI.Model.Implementation/Wallet/TransactionPreview.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public WalletId WalletId { get; } = walletId;
+    public long Amount { get; } = amount;
+    public string Address { get; } = address;
+    public IWalletAppService WalletAppService { get; } = walletAppService;
+    public long TotalFee { get; set; } = fee.Value;
+
+    public Task<Result<TxId>> Accept()
+    {
+        return WalletAppService.SendAmount(WalletId, new Amount(Amount), new Address(Address), new DomainFeeRate(feeRate)).Map(id => id);
```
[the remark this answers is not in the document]
SuperJMN: You've raised a good point, David. I agree that the name "TransactionPreview" suggests something that should only show information rather than perform actions.
I've just changed the Accept() method to Submit() which better communicates what's actually happening - sending a transaction to the blockchain:

```csharp
public interface ITransactionPreview
{
    public long TotalFee { get; set; }
    Task<Result<TxId>> Submit();
}
```

This reflects the intended UI flow: the user first sees a preview of the transaction (with fee information) and then chooses to submit it if they're satisfied.
We could also consider renaming the entire interface to something like _IPendingTransaction_ or _ITransactionDraft_ if you think that would better represent its dual purpose of preview and submission. What do you think?

src/Angor/Avalonia/Angor.UI.Model.Implementation/Wallet/TransactionPreview.cs
```diff
@@ -0,0 +1,19 @@
+using Angor.Wallet.Application;
+using Angor.Wallet.Domain;
+using CSharpFunctionalExtensions;
+
+namespace Angor.UI.Model.Implementation.Wallet;
+
+public class TransactionPreview(WalletId walletId, long amount, string address, long feeRate, Fee fee, IWalletAppService walletAppService) : ITransactionPreview
```
DavidGershony, line 7: This is weird, I can still see it as transaction Preview

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/SensitiveWalletDataProvider.cs
```diff
@@ -1,22 +1,36 @@
...
     }
-    
-    public async Task<Result<(string seed, string passphrase)>> RequestSensitiveData(WalletId id)
+
+    private async Task<Result<(string seed, string passphrase)>> RequestSensitiveDataCore(WalletId id)
     {
         return await GetEncryptedWallet(id)
-            .Bind(encryptedWallet => 
+            .Bind(encryptedWallet =>
                 walletSecurityContext.EncryptionKeyProvider.Get(id).ToResult("Encryption key not provided")
```
DavidGershony, line 34: IMHO there are too many methods here called without breaking line so it becomes really hard to see the whole flow, especially when some are called internally, I would prefer break line for each method call or add parameters and not chain everything..

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/TransactionWatcher.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Angor.Wallet.Infrastructure.Impl
+{
+    public class TransactionWatcher(IWalletAppService walletAppService) : ITransactionWatcher
+    {
+        public IObservable<Result<BroadcastedTransaction>> Watch(WalletId id)
+        {
+            var initial = Observable.Defer(() => GetTransactions(id));
+            
+            var polling = Observable.Timer(TimeSpan.FromSeconds(10))
```
DavidGershony, line 14: There is no point calling for new transactions if we are on the same block, this is also a lot of work in the wallet app service, we could check what is the block hash of the tip and check if it is different to ours before calling getTransactions so it would only refresh every 10 minutes (or 1 minute in signet etc.)
But that might be an overkill as well, perhaps there should be a trigger on the UI like when I make an investment etc. to start pulling until we get the expected change - otherwise I think the user can just refresh manually if something is missing.
What do you think?

src/Angor/Avalonia/Angor.Wallet/Infrastructure/WalletServices.cs
```diff
@@ -31,6 +36,7 @@ public static ServiceCollection Register(ServiceCollection services, ILogger log
         services.AddSingleton<ISensitiveWalletDataProvider, SensitiveWalletDataProvider>();
         services.AddSingleton<IWalletStore, WalletStore>();
         services.AddHttpClient();
+        services.AddSingleton<ITransactionWatcher, TransactionWatcher>();
```
DavidGershony, line 39: I still think there should be a specific reason to opt to singleton and otherwise we should use transient :)
lets have a talk about this when we have a meeting

src/Angor/Avalonia/Angor.UI.Model.Implementation/Wallet/TransactionPreview.cs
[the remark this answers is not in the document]
SuperJMN: Done!

src/Angor/Avalonia/Angor.UI.Model.Implementation/Wallet/TransactionPreview.cs
SuperJMN: Somewhat, I forgot to push the commits...

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/SensitiveWalletDataProvider.cs
SuperJMN: Done!

src/Angor/Avalonia/Angor.Wallet/Infrastructure/Impl/TransactionWatcher.cs
SuperJMN: We can implement any trigger later to avoid unnecessary call to the indexer. Your suggestion sounds great, by the way!

PR #281, opened 2025-03-11
Improve sidebar menu

miladsoft:


src/Angor/Client/Shared/NavMenuState.cs
```diff
@@ -1,9 +1,69 @@
...
+            ActivePage = page;
+            NotifyStateChanged();
+            await PersistState();
+        }
+
+        public async Task InitializeFromStorage()
+        {
+            try
+            {
+                var storedPage = await _jsRuntime.InvokeAsync<string>("sessionStorage.getItem", StorageKey);
```
dangershony, line 42: what are we persisting here?

src/Angor/Client/Program.cs
```diff
@@ -67,4 +67,10 @@
 
 builder.Services.AddScoped<IWalletUIService, WalletUIService>();
 
-await builder.Build().RunAsync();
\ No newline at end of file
+var app = builder.Build();
+
+// Initialize NavMenuState from storage
+var navMenuState = app.Services.GetRequiredService<NavMenuState>();
```
dangershony, line 73: we keep in storage the last page we landed on?

src/Angor/Client/Shared/NavMenuState.cs
```diff
@@ -1,9 +1,69 @@
...
 {
-    public class NavMenuState 
+    public class NavMenuState
     {
+        private readonly IJSRuntime _jsRuntime;
+        private const string StorageKey = "activeMenuPage";
+
+        public NavMenuState(IJSRuntime jsRuntime)
+        {
+            _jsRuntime = jsRuntime;
```
dangershony, line 12: you can access storage without the js we have a component for that

src/Angor/Client/Shared/NavMenuState.cs
```diff
@@ -1,9 +1,69 @@
...
+                if (_activePage != value)
+                {
+                    _activePage = value;
+                    _ = PersistState();
+                }
+            }
+        }
+
+        public async Task SetActivePage(string page)
+        {
```
dangershony, line 32: Great way to solve this

dangershony (approved):

src/Angor/Client/Shared/NavMenuState.cs
```diff
@@ -1,9 +1,72 @@
...
+        private string _activePage = string.Empty;
+        public string ActivePage
+        {
+            get => _activePage;
+            private set
+            {
+                if (_activePage != value)
+                {
+                    _activePage = value;
+                    PersistState();
```
dangershony, line 26: hmm I am not sure how I fee about pushing to storage every time we change a page, do we need to do that? anyway each page will set itself when it is loaded, can you explain again why do we need to persist  the page?

src/Angor/Client/Shared/NavMenuState.cs
```diff
@@ -1,9 +1,72 @@
...
+                }
+            }
+        }
+
+        public Task SetActivePage(string page)
+        {
+            ActivePage = page;
+            NotifyStateChanged();
+            PersistState();
```
dangershony, line 35: isn't this a duplicate? we call PersistState when we call `ActivePage = page`

miladsoft:
To enhance the user experience, the active menu tab should always be highlighted. However, some pages like Project Investment and Recover Funds are not in the menu, making navigation unclear. By storing the last active menu in the session, we ensure the correct tab remains highlighted, maintaining context and clarity for the user.

PR #290, opened 2025-03-20
Fix Project Invest issues

SuperJMN:
Improve the project invest flow to handle wallet unlocking.

Now, it's designed to **not depend on IWallet**, that can introduce some complexities I'd like to avoid.

### DIFF
--- a/src/Angor/Avalonia/AngorApp/Sections/Browse/Details/Invest/Amount/AmountViewModel.cs
+++ b/src/Angor/Avalonia/AngorApp/Sections/Browse/Details/Invest/Amount/AmountViewModel.cs
@@ -1,8 +1,9 @@
-using System.Reactive.Linq;
-using Angor.UI.Model;
+using Angor.Wallet.Application;
+using Angor.Wallet.Domain;
 using ReactiveUI.SourceGenerators;
 using ReactiveUI.Validation.Extensions;
 using ReactiveUI.Validation.Helpers;
+using Zafiro.CSharpFunctionalExtensions;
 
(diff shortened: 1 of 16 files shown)


src/Angor/Avalonia/AngorApp/Sections/Browse/Details/Invest/Amount/AmountViewModel.cs
```diff
@@ -1,29 +1,34 @@
...
 
 public partial class AmountViewModel : ReactiveValidationObject, IAmountViewModel
 {
     [Reactive] private long? amount;
 
-    public AmountViewModel(IWallet wallet, IProject project)
+    public AmountViewModel(WalletId walletId, IWalletAppService walletAppService, IProject project)
     {
-        walletBalanceHelper = wallet.Balance.ToProperty(this, model => model.WalletBalance);
+        LoadBalance = ReactiveCommand.CreateFromObservable(() => Observable.FromAsync(() => walletAppService.GetBalance(walletId)).Successes().Select(x => x.Value));
```
dangershony, line 16: Hi J, check out this new service method I made for wallet refresh, it might be best to create something similar on the shared project?

https://github.com/block-core/angor/blob/main/src/Angor/Client/Services/WalletUIService.cs#L41

SuperJMN: Nice! I'll take a look and implement it in a new PR. Thanks for the tip!

dangershony (approved):

PR #294, opened 2025-03-25
Add custom fee rate option to transactions

miladsoft:


src/Angor/Client/Shared/FeeSelector.razor
```diff
@@ -0,0 +1,191 @@
...
+                if (UseCustomFee)
+                {
+                    CustomFeeRate = FeeRate;
+                    await CustomFeeRateChanged();
+                }
+            }
+        }
+        catch (Exception ex)
+        {
+            Logger.LogError(ex, "Error setting fee position to {Position}", position);
```
dangershony, line 148: I wonder if we should show a message to the user in such errors

src/Angor/Client/Shared/FeeSelector.razor
```diff
@@ -0,0 +1,191 @@
...
+        try
+        {
+            if (UseCustomFee && CustomFeeRate > 0)
+            {
+                await OnCustomFeeRateChanged.InvokeAsync(CustomFeeRate);
+            }
+        }
+        catch (Exception ex)
+        {
+            Logger.LogError(ex, "Error changing custom fee rate to {Rate}", CustomFeeRate);
```
dangershony, line 163: same here

src/Angor/Client/Pages/Wallet.razor
```diff
@@ -1901,4 +1877,23 @@ else
...
+    
+    private long GetRecommendedMinFee()
+    {
+        // Return a recommended minimum fee rate based on network conditions
+        return FeeEstimations.Fees.Any() 
+            ? Math.Max(1, FeeEstimations.Fees.Min(f => f.FeeRate)) 
+            : 5;
+    }
+    
+    private async Task CustomFeeRateChanged(long customFeeRate)
```
dangershony, line 1889: I think I would rather have one method that takes the fee and recalculates it, and then the changed event would call that method, and this can be the pattern everywhere else.
so basically what I am saying is the method `FeeRangeChanged` should call `CustomFeeRateChanged` however maybe we should rename it to `ReCalculateFee`

src/Angor/Client/Pages/Wallet.razor
```diff
@@ -1888,17 +1866,25 @@ else
...
     }
-    
+
     protected override void OnInitialized()
     {
         NavMenuState.SetActivePage("wallet");
         base.OnInitialized();
     }
+
+    private long GetRecommendedMinFee()
```
dangershony, line 1883: what will this do exactly? do we need it?

miladsoft: This method has existed previously and is responsible for returning a recommended minimum fee rate based on current network conditions.

dangershony: ah so this is basically taking the result form the api call and presenting a default fee of the network right now.
yeah ok we defo need this

src/Angor/Client/Pages/Create.razor
```diff
@@ -1199,6 +1201,44 @@
         }
     }
 
+    private async Task RecalculateFee(decimal customFeeRate)
```
dangershony, line 1204: The method `FeeRangeChanged` should call this method otherwise we have duplicated code

src/Angor/Client/Pages/Create.razor
```diff
@@ -1199,6 +1201,44 @@
...
+
+        try
+        {
+            var words = await passwordComponent.GetWalletAsync();
+
+            // Create a custom fee estimation with the provided rate
+            var customFeeEstimation = new FeeEstimation
+                {
+                    FeeRate = (int)customFeeRate,
+                    Confirmations = 1 // Custom fees typically assume fast confirmation
```
dangershony, line 1221: we cannot assume this, what does this matter what number we put here? is it only presentation?

src/Angor/Client/Pages/Recover.razor
```diff
@@ -1173,34 +1221,58 @@
...
 
-                    feeData.SelectedFeeEstimation = feeData.FeeEstimations.Fees.OrderBy(fee => fee.Confirmations).ToList()[res - 1];
+        try
+        {
+            var words = await passwordComponent.GetWalletAsync();
 
-                    var accountInfo = storage.GetAccountInfo(network.Name);
-                    var words = await passwordComponent.GetWalletAsync();
+            // Create a custom fee estimation with the provided rate and confirmations
+            var customFeeEstimation = new FeeEstimation
```
dangershony, line 1251: why not just pass the `FeeEstimation` as a parameter instead of decimal and int?

src/Angor/Client/Pages/Spend.razor
```diff
@@ -880,34 +870,57 @@
...
-                if (res <= feeData.FeeEstimations.Fees.Count)
+            feeData.FeePosition = res;
+
+            var selectedFeeEstimation = feeData.FeeEstimations.Fees.OrderBy(fee => fee.Confirmations).ToList()[res - 1];
+
+            await RecalculateFee(selectedFeeEstimation.FeeRate, selectedFeeEstimation.Confirmations);
+        }
+    }
+
+    private async Task RecalculateFee(decimal customFeeRate, int confirmations = 0)
```
dangershony, line 883: why decimal isn't the fee represented as a long?

src/Angor/Client/Pages/Spend.razor
```diff
@@ -880,34 +870,57 @@
...
+
+            var customFeeEstimation = new FeeEstimation
                 {
-                    if (!passwordComponent.HasPassword())
-                    {
-                        showCreateModal = false;
-                        notificationComponent.ShowErrorMessage("Wallet password expired");
-                        return;
-                    }
+                    FeeRate = (int)customFeeRate,
```
dangershony, line 898: you are converting a decimal to int where it is represented as a long, it will break the fee estimation

src/Angor/Client/Pages/Spend.razor
miladsoft: It was a mistake, and I’ve fixed it. Thanks for pointing it out!

src/Angor/Client/Pages/Spend.razor
miladsoft: You're right, that was a mistake. I've corrected it to ensure the fee estimation works properly. Thanks for the review!

src/Angor/Client/Shared/FeeSelector.razor
```diff
@@ -0,0 +1,191 @@
...
+    private async Task SetFeePosition(int position)
+    {
+        try
+        {
+            if (position >= FeeMin && position <= FeeMax)
+            {
+                FeePosition = position;
+                await OnFeePositionChanged.InvokeAsync(position);
+
+                if (UseCustomFee)
```
dangershony, line 139: we invoke the `OnFeePositionChanged` and then we invoke the `OnCustomFeeRateChanged`?
this will cause the code to sign the trx twice it seems to me

src/Angor/Client/Pages/Create.razor
```diff
@@ -1172,30 +1174,50 @@
...
+            var selectedFeeEstimation = feeData.FeeEstimations.Fees.OrderBy(fee => fee.Confirmations).ToList()[res - 1];
 
-                    var accountInfo = storage.GetAccountInfo(network.Name);
-                    var unconfirmedInfo = _cacheStorage.GetUnconfirmedInboundFunds();
+            await RecalculateFee(selectedFeeEstimation.FeeRate, selectedFeeEstimation.Confirmations);
+        }
+    }
 
-                    signedTransaction = _WalletOperations.AddInputsAndSignTransaction(accountInfo.GetNextChangeReceiveAddress(), unsignedTransaction, words, accountInfo, feeData.SelectedFeeEstimation.FeeRate);
+    private async Task RecalculateFee(long customFeeRate, int confirmations = 0)
```
dangershony, line 1187: Milad I realize we can simplify this much more, instead of two methods we will make one method that will take a class that has a Position or a customFee (and maybe a bool that says useCustomFee.

Then we don't need to have two methods instead we just have one method called `RecalculateFee`
And we also dont need two events on the ui component.
this simplifies the code.

you can even call that recalcualtefee method form the send method and save the duplicate code.

dangershony (changes requested):
Please implement the idea of using one method

miladsoft:
The fee selector component now has a much cleaner architecture with:

A centralized RecalculateFee method that handles all fee calculation scenarios
A consolidated approach using the FeeCalculation class to encapsulate fee-related data
Simplified event handling through a single event callback
Better error handling with proper logging
This implementation successfully reduces code duplication while maintaining all the functionality. The component now properly handles both preset fee positions (Economy, Standard, Priority) and custom fee rates in a more maintainable way.

The UI remains user-friendly while the underlying code is more robust.

src/Angor/Client/Pages/Invest.razor
```diff
@@ -781,34 +768,46 @@ else
...
+                feeData.SelectedFeeEstimation.FeeRate);
 
-                    StateHasChanged();
-                }
-            }
+            StateHasChanged();
+        }
+        catch (Exception ex)
+        {
+            notificationComponent.ShowErrorMessage("Error calculating fee: " + ex.Message);
```
dangershony, line 810: why do we handle the error differently here and in create?

src/Angor/Client/Pages/Wallet.razor
```diff
@@ -1607,29 +1579,29 @@ else
 
     private async Task FeeRangeChanged(ChangeEventArgs e)
     {
-        // Here you can handle the fee range change
-        // The new value is in e.Value
-
         var selected = e.Value?.ToString();
 
-        if (selected != null)
+        if (selected != null && int.TryParse(selected, out var position))
```
dangershony, line 1584: this method is not called anymore, I will delete it

src/Angor/Client/Pages/Wallet.razor
```diff
@@ -1607,29 +1579,29 @@ else
...
+                var estimationsFee = FeeEstimations.Fees.OrderBy(fee => fee.Confirmations).ToList()[position - 1];
+                await RecalculateFee(estimationsFee.FeeRate, estimationsFee.Confirmations);
+            }
+        }
+    }
 
-                    _sendInfo.FeeBlockCount = estimationsFee.Confirmations;
-                    _sendInfo.FeeRate = estimationsFee.FeeRate;
-                    var accountInfo = storage.GetAccountInfo(network.Name);
+    private async Task RecalculateFee(long feeRate, int blockCount = 0)
```
dangershony, line 1596: also this method is not called

PR #295, opened 2025-03-26
For elecrs not all data is indexed so we need to make extra calls to …

dangershony:
…the indexer

### DIFF
--- a/src/Angor/Shared/Services/MempoolSpaceIndexerApi.cs
+++ b/src/Angor/Shared/Services/MempoolSpaceIndexerApi.cs
@@ -16,8 +16,8 @@ public class MempoolSpaceIndexerApi : IIndexerService
     private readonly HttpClient _httpClient;
     private readonly INetworkService _networkService;
 
-    private const string AngorApiRoute = "/api/V1/query/Angor";
-    private const string MempoolApiRoute = "/api/V1";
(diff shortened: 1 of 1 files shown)


src/Angor/Shared/Services/MempoolSpaceIndexerApi.cs
```diff
@@ -414,6 +416,53 @@ public async Task<string> GetTransactionHexByIdAsync(string transactionId)
...
+                        _networkService.CheckAndHandleError(response);
+
+                        if (!response.IsSuccessStatusCode)
+                            throw new InvalidOperationException(response.ReasonPhrase);
+
+                        var trx = await response.Content.ReadFromJsonAsync<List<MempoolTransaction>>(new JsonSerializerOptions()
+                            { PropertyNamingPolicy = JsonNamingPolicy.SnakeCaseLower });
+
+                        bool found = false;
+                        foreach (var transaction in trx)
```
DavidGershony, line 444: I think it's better to first find the transaction and than iterate through the inputs in it, you go though thousands of transactions and you go through every vin in them before reaching your data for no reason.

dangershony: ideally yes but I realize the trx id to compare against is inside the input so I anyway have to iterate each vin inside each trx.

PR #297, opened 2025-03-27
Add website status check for explorer URLs and improve error logging

miladsoft:


src/Angor/Shared/Services/NetworkService.cs
```diff
@@ -137,18 +138,23 @@ public async Task CheckServices(bool force = false)
 
                     try
                     {
-                        var uri = new Uri(explorerUrl.Url);
-                        
-                         var heartbeatUrl = new Uri(uri, "api/stats/heartbeat").ToString();
-                         var response = await _httpClient.GetAsync(heartbeatUrl);
+                        string apiUrl = $"https://urlstatus.netlify.app/api/status?url={explorerUrl.Url}";
```
dangershony, line 141: this is some external dependency

dangershony:
As discussed we won't use an external service to check the status of an explorer, in fact I think we can remove the explorer status indicator and just allow to add an explorer link.

PR #298, opened 2025-03-27
resolve the spend destination

dangershony:


src/Angor/Client/Pages/Spend.razor
```diff
@@ -781,12 +801,49 @@
...
+                    }
 
                     continue;
                 }
 
-                item.IsSpent = LinqExtension.Contains(unconfirmedOutbound, new Outpoint(item.Trxid, item.Outputindex));
+                item.IsSpent = unconfirmedOutbound.ContainsOutpoint(new Outpoint(item.Trxid, item.Outputindex));
+                if (item.IsSpent)
+                {
+                    item.SpentType = "pending";
```
DavidGershony, line 845: Can you add a comment so it's clear what pending means? is it not related to who spent it?

src/Angor/Client/Pages/Spend.razor
```diff
@@ -781,12 +801,49 @@
...
+
+                    // try to resolve the destination
+                    var spentInTransaction = await _IndexerService.GetTransactionInfoByIdAsync(output.SpentInTransaction);
+
+                    var input = spentInTransaction?.Inputs.FirstOrDefault(input => input.InputTransactionId == item.Trxid && input.InputIndex == item.Outputindex);
+
+                    var investmentTransaction = trxsHexCache.TryGet(item.Trxid);
+                    if (investmentTransaction == null)
+                    {
+                        investmentTransaction = network.CreateTransaction(await _IndexerService.GetTransactionHexByIdAsync(item.Trxid));
```
DavidGershony, line 814: please move await _IndexerService.GetTransactionHexByIdAsync(item.Trxid) to a different line

src/Angor/Client/Pages/Spend.razor
dangershony: To solve who spent pending we would need to fetch data of unconfirmed trx, it may be possible but I think it is best to do it in a separate pr because we dont really try to resolve pending trx anywhere

PR #304, opened 2025-03-29
Disabling animation in sidenav and auto hiding

### LINKED ISSUE #199: Menu - Disable Animation/Auto-Hide
Could perhaps the side-menu be more basic, when it is closed it is closed with only icons, or expanded it keeps open and doesn't require the "Lock" to be activated? The way it works now is a bit annoying, which is why I was inspired to make this issue 👍

vivekmaurya001:
resolves #199

Currently the navbar is too much complicated with very random expansion and compressed nature, Also in mobile view there are three buttons just for closing the navbar,   The cross at Top left , the arrow on bottom left and hamburger icon :-
image
 

previously it look like :-
[Screencast from 29-03-25 02:45:18 PM IST.webm](https://github.com/user-attachments/assets/5e6a6200-919f-418b-b1c2-60a871f2457a)

After changes :- 
[Screencast from 29-03-25 02:47:55 PM IST.webm](https://github.com/user-attachments/assets/b41cd55c-edbe-4c27-90ab-9a5d3eee2de1)


miladsoft:
According to the discussion on Discord by the creator of this PR, it was decided that the menu would be completely changed in the new PR.

Issue #211, opened 2024-12-12
After wallet restore, first opening the "founder" page should perform a rescan automatically


sondreb:
We should store some local store key and ensure that rescan is performed on the founder page so users don't need to manually scan after restoring wallet.

Keerthi421:
Pull Request Description
This PR resolves Issue #211 by ensuring that after a wallet is restored, the first time the user opens the "Founder" page, an automatic rescan is triggered.

Key Changes:
Added a rescan flag (needs_rescan) in WalletStorage to track when a rescan is needed.
Modified SaveWalletWords() to set the rescan flag when a wallet is restored or created.
Updated OnInitializedAsync() in Founder.razor to:
Check if a rescan is needed on page load.
Perform the rescan automatically if required.
Clear the rescan flag after completion.

dangershony:
This is not needed the rescan happens on create, and the user can just press on the scan if they want

PR #310, opened 2025-04-02
Automatically Perform Rescan After Wallet Restore with UI Feedback & …

Keerthi421:
Pull Request Description
This PR resolves Issue https://github.com/block-core/angor/issues/211 by ensuring that after a wallet is restored, the first time the user opens the "Founder" page, an automatic rescan is triggered.

Key Changes:
Added a rescan flag (needs_rescan) in WalletStorage to track when a rescan is needed.
Modified SaveWalletWords() to set the rescan flag when a wallet is restored or created.
Updated OnInitializedAsync() in Founder.razor to:
Check if a rescan is needed on page load.
Perform the rescan automatically if required.
Clear the rescan flag after completion.

### DIFF
--- a/src/Angor/Client/Pages/Founder.razor
+++ b/src/Angor/Client/Pages/Founder.razor
@@ -5,6 +5,8 @@
 @using Angor.Shared.Services
 @using Nostr.Client.Messages
 @using Angor.Client.Shared
+@using System.Threading
+@using System.Threading.Tasks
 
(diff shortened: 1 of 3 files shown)


dangershony:
I think we already do the project scanning on the founder page when we create a new project.

dangershony:
This pr is not needed anymore

PR #312, opened 2025-04-03
Fix: Disable animation & auto-hide in sidebar menu #199

### LINKED ISSUE #199: Menu - Disable Animation/Auto-Hide

rajanarahul93:
###  Changes Made:
1. **Removed Auto-Hide Behavior**  
   - Deleted the `pointerover` styles that caused the sidebar to expand/collapse on hover.

2. **Removed Expand/Collapse Animation**  
   - Updated the `controls|Pane` style to always keep the sidebar at `ExpandedWidth`.
   - Removed `DoubleTransition` animation for width changes.

###  Issue Fixed:
- The sidebar now **remains expanded** and **does not collapse on hover**.

resolves #199 
@sondreb @dangershony

### DIFF
--- a/src/Angor/Avalonia/AngorApp/Sections/Shell/Sidebar.axaml
+++ b/src/Angor/Avalonia/AngorApp/Sections/Shell/Sidebar.axaml
@@ -32,15 +32,8 @@
             <Setter Property="Height" Value="{StaticResource IconSize}" />
         </Style>
         <Style Selector="controls|Pane">
-            <Setter Property="Width" Value="{StaticResource CollapsedWidth}" />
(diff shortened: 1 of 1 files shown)


miladsoft:
https://github.com/block-core/angor/issues/199

Sondre meant the Blazor project, and you made changes to the Avalonia project, which is still in the early stages of design.

dangershony:
@rajanarahul93 I am closing this as it is on the wrong framework

PR #320, opened 2025-04-05
Redesigned sidebar

vivekmaurya001:
Completed the sidebar with new design

https://github.com/user-attachments/assets/b6daf881-2b29-493d-822d-ad5ea6a78e2c

cc: @miladsoft @dangershony


miladsoft:
image

The footer should be positioned at the bottom of the page. However, when the content of the page is long, the footer should not stick to the bottom of the screen. Instead, it should appear after all the content. In such cases, the user should be able to scroll down the page and see the footer only after scrolling past the content.

image

src/Angor/Client/Shared/MainLayout.razor.css
```diff
@@ -0,0 +1,23 @@
+.main-content {
+    position: relative;
+    margin-left: 300px;
```
miladsoft, line 3: margin-left: 290px;

miladsoft:
image
The icons must be exactly centered both vertically and horizontally. This means they should be placed in the exact middle from top to bottom and from left to right.

miladsoft:
image
The background of this button, on hover, must be exactly the same as the other menu items. It should have the same color and match the style of the rest of the menu when hovered over.

miladsoft:
image
The logo must be exactly centered. It should be placed in the middle both vertically and horizontally. Additionally, it should have an animated effect on hover, just like it had before.

src/Angor/Client/Shared/MainLayout.razor.css
miladsoft: fixed

dangershony (approved):

PR #321, opened 2025-04-06
Caching nostr dms

sdivjot:
Added methods in relay service and client storage to fetch and store the nostr dms in local storage using blazored storage, and changed the dm lookups to fetch old from local storage, and fetch new ones using since parameter

### DIFF
--- a/src/Angor/Shared/Services/RelayService.cs
+++ b/src/Angor/Shared/Services/RelayService.cs
@@ -11,30 +11,38 @@ namespace Angor.Shared.Services
 {
     public class RelayService : IRelayService
     {
-        private ILogger<RelayService> _logger;
+        private readonly ILogger<RelayService> _logger;
         private readonly INostrCommunicationFactory _communicationFactory;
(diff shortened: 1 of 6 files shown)


dangershony:
Thanks, due to limited time it might take me so e time to review this

DavidGershony (approved):

src/Angor/Shared/Services/RelayService.cs
```diff
@@ -96,65 +104,130 @@ public Task LookupSignaturesDirectMessagesForPubKeyAsync(string nostrPubKey, Dat
 
             var subscriptionKey = nostrPubKey + "DM";
 
+            // Process cached events first
+            var cachedEvents = _nostrEventCacheService.GetCachedEvents(subscriptionKey);
```
dangershony, line 108: I just wonder what efficiency gains do we get here? we anyway create a subscription and call the relay.
Events will be fired twice once from cache and then again from each relay.

dangershony:
please fix the conflicts

Issue #247, opened 2025-01-14
Use the new NIP17 for encryption on DMs


dangershony:
Right now we use [NIP4](https://github.com/nostr-protocol/nips/blob/master/04.md) for sending encrypted signatures between founder and investor, that NIP is deprecated we must instead use [NIP17](https://github.com/nostr-protocol/nips/blob/master/17.md) and [NIP44](https://github.com/nostr-protocol/nips/blob/master/44.md) / [NIP59](https://github.com/nostr-protocol/nips/blob/master/59.md)

PR #333, opened 2025-04-19
Feature: Using the new Nip17 for encryption on DMs

ritankarsaha:
This PR is aimed at fixing issue #247 

## Description :-
This PR replaces the old NIP‑04–based encryption flow with the newer NIP‑17/NIP‑44/NIP‑59 specs by:
- Bundling nostr-tools in the browser
- Adding nostrInterop.js wrappers (window.nip17Encrypt/Decrypt, nip59Encrypt/Decrypt, etc.)
- Updating index.html to load those modules
- Wiring the Blazor EncryptionService to call the new JS methods in place of the deprecated NIP‑04 APIs.

## Steps to Reproduce :- 
1. Apply the patch
2. Hop onto Angor
3. Inspect -> In the networks tab reload the page to ensure the new js component that is added gets loaded properly and shows a 200.
4. Move onto the console tab

```js
console.log("nip57Encrypt:", typeof window.nip57Encrypt);
console.log("nip57Decrypt:", typeof window.nip57Decrypt);
console.log("nip17Encrypt:", typeof window.nip17Encrypt);
console.log("nip59Encrypt:", typeof window.nip59Encrypt);
console.log("nip59Decrypt:", typeof window.nip59Decrypt);
```

Type this out and hit enter. You should get output like -- 

```js
nip57Encrypt: function
nip57Decrypt: function
nip17Encrypt: function
nip59Encrypt: function
nip59Decrypt: function
```

This shows these are getting defined properly .

 
Screenshot 2025-04-19 at 4 04 38 PM

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -3,17 +3,6 @@
 version = 3
 
 [[package]]
-name = "angor"
-version = "0.0.83"
-dependencies = [
- "serde",
- "serde_json",
- "tauri",
- "tauri-build",
- "tauri-plugin-shell",
-]
-
-[[package]]
 name = "addr2line"
 version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -65,6 +54,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
 dependencies = [
  "libc",
+]
+
+[[package]]
+name = "angor"
+version = "0.0.83"
+dependencies = [
+ "serde",
+ "serde_json",
+ "tauri",
+ "tauri-build",
+ "tauri-plugin-shell",
 ]
 
 [[package]]
--- a/src/Angor/Client/Services/EncryptionService.cs
+++ b/src/Angor/Client/Services/EncryptionService.cs
@@ -1,4 +1,5 @@
-﻿using Blockcore.NBitcoin;
+﻿using System.Threading.Tasks;
+using Blockcore.NBitcoin;
 using Blockcore.NBitcoin.Crypto;
 using Blockcore.NBitcoin.DataEncoders;
 using Microsoft.JSInterop;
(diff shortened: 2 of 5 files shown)


src/Angor/Client/Services/EncryptionService.cs
```diff
@@ -8,42 +9,38 @@ namespace Angor.Client.Services
...
         {
-            _jsRuntime = jsRuntime;
+            _js = jsRuntime;
         }
 
-        public async Task<string> EncryptData(string secretData, string password)
-        {
-            return await _jsRuntime.InvokeAsync<string>("encryptData", secretData, password);
-        }
+        public Task<string> EncryptData(string secretData, string password)
```
itailiors, line 19: Same thing over here, try to keep changes to formatting as little as possible (unless you think they are really needed and its unreadable)
makes it difficult to go over PR's like this,
like a smart man @dangershony once said 
"We welcome contributions to the codebase. Please maintain the same code style as the existing code."

src/Angor/Client/Services/EncryptionService.cs
```diff
@@ -8,42 +9,38 @@ namespace Angor.Client.Services
...
 
-        private static string GetSharedSecretHexWithoutPrefix(string nsec, string npub)
+        private static string GetSharedSecretHex(string nsec, string npub)
         {
-            var privateKey = new Key(Encoders.Hex.DecodeData(nsec));
-            var publicKey = new PubKey("02" + npub);
-            
-            var secert = publicKey.GetSharedPubkey(privateKey);
-            return Encoders.Hex.EncodeData(secert.ToBytes()[1..]);
+            var priv = new Key(Encoders.Hex.DecodeData(nsec));
```
itailiors, line 39: why change the variable names?

src/Angor/Client/wwwroot/index.html
```diff
@@ -57,8 +57,13 @@
             </div>
         </div>
     </div>
+    <script type="module">
```
itailiors, line 60: this also seems unorthodox,
you’re directly importing the nostr-tools library from a cdn using an inline <script> tag. 
while this works, it might be better to align with the existing structure by creating a dedicated module file for this import, similar to how other scripts like service-crypto.js and nostrInterop.js are

src-tauri/Cargo.lock
```diff
@@ -67,6 +56,17 @@ dependencies = [
  "libc",
 ]
 
+[[package]]
```
itailiors, line 59: and also im not a tauri master,
but then not sure if this should be in the .gitignore or not 
seems harmless, but not sure why it's here?

ritankarsaha: Apologies, these change has been carried on when i made this branch from the main. It got un-noticed by me. i will delete this commit `cargo file changes` in the interactive rebase.

dangershony:
I will need @DavidGershony and @sondreb to review this

dangershony:
What about backwards compatibility? (not sure we need that though)

src-tauri/Cargo.lock
ritankarsaha: @itailiors dropped this commit.

ritankarsaha:
> What about backwards compatibility? (not sure we need that though)

@dangershony We can like try the new NIP‑17/NIP‑59 flow first for the older things, then fall back to NIP‑04 if decryption fails ??

src/Angor/Client/wwwroot/index.html
ritankarsaha: okaay @itailiors working on this and making the necessary tweaks soon :) !!

PR #216, opened 2024-12-13
Upgrade to nbitcoin 7.0.46

dangershony:
This is a draft till we are sure that we fixed all the issues with nbitcoin

PR #335, opened 2025-04-19
fix : updated dotnet version to 9

vkprogrammer-001:
- Updated the .NET version to 9. As per the contribution.md


vkprogrammer-001:
Hi @sondreb, can you please review this PR? Thank you!

sondreb (changes requested):
This will not be merged until we have tested it properly. Last time we upgraded we got many random different errors. We need to deploy a separate test instance of the app and run it for days with testing to ensure no issues occurs. Thanks for the PR, we'll leave it open.

dangershony:
As @sondreb said there is a PR pending for this but, upgrading to dotnet9 breaks the taproot scripts
https://github.com/block-core/angor/pull/216

@vkprogrammer-001 have you tested the whole flow before submitting this PR?

vkprogrammer-001:
@dangershony 
No, I just built and ran it using .NET. It works fine there, but I haven't tested the whole flow yet.

vkprogrammer-001:
Thank you for pointing that out. I'll ensure to thoroughly test the UI manually since the current tests focus on the core code and not the UI. Let me know if there are specific scenarios or areas you'd like me to prioritize during the manual testing process.

PR #350, opened 2025-04-28
Add Invest to Project to Avalonia App

SuperJMN:
Adds the actual Invest feature

Follow-up PR from https://github.com/block-core/angor/pull/299


src/Angor/Avalonia/Angor.Contexts.CrossCutting/Angor.Contexts.CrossCutting.csproj
```diff
@@ -11,6 +11,7 @@
     <ItemGroup>
         <PackageReference Include="AspectInjector" />
         <PackageReference Include="CSharpFunctionalExtensions" />
+        <PackageReference Include="JetBrains.Annotations" />
```
DavidGershony, line 14: What is this added for?

src/Angor/Avalonia/Angor.Contexts.Wallet/Domain/BroadcastedTransactionExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+    
+    public static Amount GetTotalSent(this BroadcastedTransaction tx)
+        => new Amount(tx.AllOutputs
+            .Where(o => tx.WalletOutputs.All(w => w.Address != o.Address))
+            .Sum(o => o.Amount.Sats));
+    
+    public static Amount GetTotalReceived(this BroadcastedTransaction tx)
+        => new Amount(tx.WalletOutputs.Sum(o => o.Amount.Sats));
+    
+    public static IReadOnlyCollection<Address> GetSenderAddresses(
```
DavidGershony, line 20: Could use some description here, are you only taking inputs that do not belong to the wallet?

src/Angor/Avalonia/AngorApp/Features/Invest/Amount/Breakdown.cs
```diff
@@ -0,0 +1,13 @@
+namespace AngorApp.Features.Invest.Amount;
+
+public record Breakdown(
+    int Index,
+    long Amount,
+    double Weight,
```
DavidGershony, line 6: What is the meaning of weight? the percentage of the total?

SuperJMN: Good question. Since it goes from 0 to 1, it's not a percentage, but a proportion. I was wondering about a good name initially, and used that. But it's clearly misleading.

I think "Ratio" or "Proportion" would be clearer names. What do you think? Maybe "RatioOfTotal" makes it.

src/Angor/Avalonia/Angor.Contexts.Wallet/Domain/BroadcastedTransactionExtensions.cs
SuperJMN: Yes, I'll add XML documentation to clarify these methods.

The `GetTotalSent` method calculates the total amount sent to external addresses. It uses `tx.AllOutputs` and filters them to only include outputs where the address is not in any of the wallet's own outputs (`.Where(o => tx.WalletOutputs.All(w => w.Address != o.Address))`).

This distinction is important because:
- `WalletInputs` and `WalletOutputs` represent the wallet's own transactions
- `AllInputs` and `AllOutputs` include all transaction parts, both the wallet's and external ones

I'll add proper documentation to make these calculations clearer.

src/Angor/Avalonia/Angor.Contexts.CrossCutting/Angor.Contexts.CrossCutting.csproj
SuperJMN: Ah, yes! This package reference was only being used for a single [UsedImplicitly] attribute. I agree it's unnecessary overhead for such limited usage.

I'll remove this reference completely. Thanks for pointing it out!

src/Angor/Avalonia/AngorApp/Features/Invest/Amount/Breakdown.cs
SuperJMN: I've used `RatioOfTotal`. Please, tell me if it's OK :)

DavidGershony (approved):

dangershony (approved):

PR #351, opened 2025-04-28
Add password strength requirements to wallet creation

vkprogrammer-001:
## Overview
This PR adds password strength validation during wallet creation to improve security and prevent users from creating wallets with weak passwords.

## Changes
- Added password validation that requires:
   - Minimum 8 characters
   - At least one uppercase letter
   - At least one lowercase letter
   - At least one number or special character
- Added real-time password strength feedback in the UI
- Added password requirements list in the wallet creation wizard
- Implemented consistent validation across both new wallet creation and wallet recovery flows

## Testing
- Verified password requirements are enforced when creating a new wallet
- Verified password requirements are enforced when recovering a wallet
- Confirmed the "Next" and "Create Wallet" buttons remain disabled until a valid password is provided
- Tested with various password combinations to ensure correct validation behavior

## Screenshots/ Recording

https://github.com/user-attachments/assets/60ffb0b1-a256-4e3b-a8a5-2a998adf0ee7

### DIFF
--- a/src/Angor/Client/Pages/Wallet.razor
+++ b/src/Angor/Client/Pages/Wallet.razor
@@ -225,6 +225,17 @@
                                                         }
                                                     </button>
                                                 </div>
+                                                <div class="@(IsPasswordValid() ? "text-success" : "text-danger") mt-2 small">
+                                                    @GetPasswordStrengthFeedback()
+                                                </div>
+                                                <div class="text-muted mt-2 small">
+                                                    <p>For a strong password:</p>
+                                                    <ul>
+                                                        <li>Use at least 8 characters</li>
+                                                        <li>Include uppercase and lowercase letters</li>
+                                                        <li>Include at least one number or special character</li>
+                                                    </ul>
+                                                </div>
                                             </div>
                                             break;
                                     }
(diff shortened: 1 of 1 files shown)


vkprogrammer-001:
@miladsoft @sondreb @dangershony, can you please review it? Thank you!

miladsoft:
On the TestNet, allow easy password entry and on the MainNet, check like this. I like it ❤️

vkprogrammer-001:
sure. I will do it.

vkprogrammer-001:
> On the TestNet, allow easy password entry and on the MainNet, check like this. I like it ❤️

@miladsoft I've implemented the requested changes.

dangershony:
Do we really want that restrictions?

dangershony:
I understand what you are saying but I dont think we need a password strength check, anyway security will come form hardware wallets and remote signets not a stronger password.

vkprogrammer-001:
May I ask a question?
> I understand what you are saying but I dont think we need a password strength check, anyway security will come form hardware wallets and remote signets not a stronger password.

I understand your point about not needing a password strength check, as security will be handled by hardware wallets and remote signets rather than a stronger password. However, could you please clarify why we believe this is unnecessary?

Screenshot 2025-04-29 at 4 01 36 PM

If a password is easily hackable, does it serve any purpose? In that case, perhaps we should reconsider using passwords altogether, as it might just be an additional step for users without providing significant security benefits.

dangershony:
A user can still put a strong password we did not take that option away, this pr only mandates strong passwords.
Passwords give an illusion of security, because this isa browser env even a strong password decrypts the rods in to the browser.
So I rather not add such a restrictions of strong password and let the user decide what they prefer.

PR #359, opened 2025-05-01
Update word verification logic to use selected word indices

miladsoft:
image


dangershony:
yeah I didn't get why it was shuffled

dangershony (approved):

PR #361, opened 2025-05-02
PSBT support - separate the add inputs and sign and use psbt

dangershony:
This is still wip but it will replace the method `AddInputsAndSignTransaction` with two methods `CreatePsbtForTransaction` and `SignPsbt`


src/Angor/Shared/Models/PsbtWrapper.cs
```diff
@@ -0,0 +1,8 @@
+using Blockcore.Consensus.TransactionInfo;
+
+namespace Angor.Shared.Models;
+
+public class PsbtWrapper
```
DavidGershony, line 5: Perhaps PsbtData or PsbtMessage?

src/Angor/Shared/WalletOperations.cs
```diff
@@ -35,7 +36,84 @@ public string GenerateWalletWords()
...
+
+        changeAddress = changeAddress ?? accountInfo.GetNextChangeReceiveAddress();
+
+        var utxoDataWithPaths = FindOutputsForTransaction((long)transaction.Outputs.Sum(_ => _.Value), accountInfo);
+        var coins = utxoDataWithPaths.Select(u => new Coin(uint256.Parse(u.UtxoData.outpoint.transactionId), (uint)u.UtxoData.outpoint.outputIndex, Money.Satoshis(u.UtxoData.value), Script.FromHex(u.UtxoData.scriptHex))).ToList();
+
+        if (!coins.Any())
+            throw new ApplicationException("No coins found to fund the transaction.");
+
+        var builder = new TransactionBuilder(network)
```
DavidGershony, line 54: Are we going to use the transaction builder after all?

src/Angor/Shared/WalletOperations.cs
```diff
@@ -35,7 +36,84 @@ public string GenerateWalletWords()
...
+            if (path != utxoInfo.HdPath)
+                throw new InvalidOperationException($"Path does not match {path} {utxoInfo.HdPath}");
+
+            psbt.Inputs[i].HDKeyPaths.Add(new NBitcoin.PubKey(pubKey.ToBytes()), rootedKeyPath);
+        }
+
+        return new PsbtWrapper { PsbtHex = psbt.ToHex() };
+    }
+
+    public TransactionInfo SignPsbt(PsbtWrapper psbtWrapper, WalletWords walletWords)
```
DavidGershony, line 92: Will we be sending the wallet words to the other wallets as well?

DavidGershony (changes requested):
Need some explenations

src/Angor/Shared/Models/PsbtWrapper.cs
dangershony: yeah I also didn't like that name

src/Angor/Shared/WalletOperations.cs
dangershony: hmm this was the easy way, I wonder if it is possible to use psbt only, I am going to try.
If we can bypass the builder it will be my preferred option.

src/Angor/Shared/WalletOperations.cs
dangershony: no but assuming this code will be called form the desktop or a signing device then the words will be available there

src/Angor/Shared/Models/AccountInfo.cs
```diff
@@ -3,6 +3,7 @@ namespace Angor.Shared.Models;
 public class AccountInfo
 {
     public string ExtPubKey { get; set; }
+    public string RootExtPubKey { get; set; }
```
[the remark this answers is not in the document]
dangershony: ExtPubKey  path is `m/84'/1'/0'`
RootExtPubKey path is `m`

the psbt needs the RootExtPubKey to work

src/Angor/Shared/WalletOperations.cs
DavidGershony: No easy way, either we use it than have a wrapper for it or we don't use it, otherwise we create more code to fix later on

src/Angor/Shared/WalletOperations.cs
DavidGershony: Perhaps we should have a different approach and inject the access to the Wallet words in our implementation and otherwise use external wallet with the same Interface

src/Angor/Shared/Models/PsbtWrapper.cs
DavidGershony: PsbtDto?

dangershony: `PsbtData` or `PsbtInfo`  are my preferences

DavidGershony (commented):
Needs some updates

src/Angor/Shared/Models/PsbtWrapper.cs
dangershony: PsbtData it is

src/Angor/Shared/WalletOperations.cs
dangershony: I dont know yet, it will be more clear when we come to implement HW

src/Angor/Shared/WalletOperations.cs
dangershony: I will wait with this decision, as it is separated to the psbt part.
I rather make sure psbt with taproot works before we start to optimize how we build trx

src/Angor/Client/Pages/Invest.razor
```diff
@@ -882,7 +882,9 @@ else
 
             unSignedTransaction = _InvestorTransactionActions.CreateInvestmentTransaction(project.ProjectInfo, InvestorPubKey, Money.Coins(Investment.InvestmentAmountBtc).Satoshi);
 
-            signedTransaction = _WalletOperations.AddInputsAndSignTransaction(accountBalanceInfo.AccountInfo.GetNextChangeReceiveAddress(), unSignedTransaction, words, accountBalanceInfo.AccountInfo, feeData.SelectedFeeEstimation.FeeRate);
+            var psbt = _WalletOperations.CreatePsbtForTransaction(unSignedTransaction, accountBalanceInfo.AccountInfo, feeData.SelectedFeeEstimation.FeeRate);
+
+            signedTransaction = _WalletOperations.SignPsbt(psbt, words);
```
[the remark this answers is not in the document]
dangershony: Agreed to do that as a separate feature

DavidGershony (approved):

PR #362, opened 2025-05-02
Replace metadata info kind from ApplicationSpecificData (30078) to 3030

dangershony:
The idea in this change is that kind 3030 is not replaceable, the metadata cannot be replaced once published 

The link to the nip description
https://github.com/block-core/nips/blob/peer-to-peer-decentralized-funding/3030.md?tab=readme-ov-file


src/Angor/Shared/Services/RelayService.cs
```diff
@@ -271,11 +271,14 @@ public Task<string> DeleteProjectAsync(string eventId, string hexPrivateKey)
...
-        private static NostrEvent GetNip78NostrEvent( string content)
+        private static NostrEvent GetNip3030NostrEvent( string content)
         {
+            // https://github.com/block-core/nips/blob/peer-to-peer-decentralized-funding/3030.md
+
             var ev = new NostrEvent
             {
-                Kind = NostrKind.ApplicationSpecificData,
+                //Kind = NostrKind.ApplicationSpecificData,
+                Kind = (NostrKind)3030,
```
DavidGershony, line 281: Did you change all the filteres on the requests as well?

dangershony: I dont think we use that as filter but I will check

DavidGershony:
This change means no old projects will work

sondreb:
> This change means no old projects will work

If the code to retrieve these events specify the kinds, this can just be removed. The events are retrieved using `id` anyway, kind does not matter (but to relays it does, since it can no longer be replaced).

dangershony:
> > This change means no old projects will work
> 
> If the code to retrieve these events specify the kinds, this can just be removed. The events are retrieved using `id` anyway, kind does not matter (but to relays it does, since it can no longer be replaced).

In a few subscriptions we specify the kind but I just added the new kind to the filter.

dangershony:
Should we publish the nip to the main nip repo?

sondreb:
> Should we publish the nip to the main nip repo?

No, don't bother.

src/Angor/Shared/Services/RelayService.cs
```diff
@@ -16,8 +16,10 @@ public class RelayService : IRelayService
         private readonly INetworkService _networkService;
         private readonly IRelaySubscriptionsHandling _subscriptionsHandling;
         private readonly ISerializer _serializer;
-        
-        
+
+        // https://github.com/block-core/nips/blob/peer-to-peer-decentralized-funding/3030.md
+        public const NostrKind Nip3030NostrKind = (NostrKind)3030;
```
DavidGershony, line 21: ProjectInfoEventKind?

DavidGershony: Once We have the NuGet package for our fork of Nostr-client we will add that enum to it

dangershony: Yes I agree

PR #367, opened 2025-05-08
Add a check box to stay logged in when the user is active

dangershony:
If they user stays active they wont need to put the password if they tick that stay logged in box

image

### DIFF
--- a/src/Angor/Client/Shared/PasswordComponent.razor
+++ b/src/Angor/Client/Shared/PasswordComponent.razor
@@ -50,7 +50,7 @@
                                 </button>
                             </div>
                         </div>
-
+                        
                         <div class="mb-3">
                             <label for="cacheDuration">Remember password for:</label>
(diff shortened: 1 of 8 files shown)


src/Angor/Client/Shared/PasswordComponent.razor
```diff
@@ -60,10 +60,19 @@
                             </select>
                         </div>
 
+                        <div class="form-check mb-3">
+                            <input class="form-check-input border mt-0" type="checkbox"
+                                   id="confirmStayLoggedIn" @bind="_passwordCacheService.StayLoggedIn">
+                            <label class="form-check-label mb-0" for="confirmStayLoggedIn">
+                                Stay Logged In?
```
DavidGershony, line 67: The term logged in is correct?

DavidGershony (changes requested):
The logged in name should change

PR #372, opened 2025-05-13
Adding feature flags

theDRB123:
Purpose:
- Adding a feature flag services to dynamically enable/disable specific features when on Angornet

Changes:
- Select network modal updated to show Feature Flag section
- Added FeatureFlag service 

image

### DIFF
--- a/src/Angor/Client/Pages/Settings.razor
+++ b/src/Angor/Client/Pages/Settings.razor
@@ -15,7 +15,7 @@
 @inject NavMenuState NavMenuState
 @inject ILogger<Settings> Logger;
 @inject NavigationManager _navManager
-
+@inject IFeatureFlagService _featureFlagService
 @inherits BaseComponent
 <NotificationComponent @ref="notificationComponent" />
 
@@ -116,7 +116,29 @@
                                             </div>
                                         </div>
                                     </div>
-
+                                    @if (selectedNetwork == "Angornet")
+                                    {
+                                        <div class="modal-body modal-body-scroll py-4">
+                                            <div class="transaction-info-section">
+                                                <div class="experimental-features mt-2">
+                                                    <h5 class="mb-3">Enable Experimental Features</h5>
+                                                    @foreach (var feature in _featureFlags)
+                                                    {
+                                                        <div class="form-check mb-3">
+                                                            <input class="form-check-input"
+                                                                type="checkbox"
+                                                                id="@feature.Key"
+                                                                checked="@feature.Value"
+                                                                @onchange="(e) => OnFeatureFlagChanged(feature.Key, e.Value is bool value && value)" />
+                                                            <label class="form-check-label" for="@feature.Key">
+                                                                @feature.Key
+                                                            </label>
+                                                        </div>
+                                                    }
+                                                </div>
+                                            </div>
+                                        </div>
+                                    }
                                     <div class="modal-footer border-0 pt-0">
                                         <button class="btn btn-border-success w-100 w-md-auto" @onclick="ChangeNetwork">
                                             Change Network
@@ -554,6 +576,7 @@
 
 
 @code {
+
     private string newIndexerLink;
     private string newRelayLink;
     private string newExplorerLink;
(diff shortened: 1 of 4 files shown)


src/Angor/Client/Services/FeatureFlagService.cs
```diff
@@ -0,0 +1,39 @@
+using Blazored.LocalStorage;
+
+namespace Angor.Client.Services;
+
+public class FeatureFlagService : IFeatureFlagService
+{
+    private readonly ISyncLocalStorageService _storage; // Injected storage service
```
dangershony, line 7: ISyncLocalStorageService is used in a dedicated storage service, don't inject it directly here instead use the interface `IClientStorage`

src/Angor/Client/Pages/Settings.razor
```diff
@@ -116,7 +116,29 @@
                                             </div>
                                         </div>
                                     </div>
-
+                                    @if (selectedNetwork == "Angornet")
```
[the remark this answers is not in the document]
theDRB123: If we are keeping these as experimental features i think we should only allow on testnets, 

I believe that feature flags for mainnet are more applicable when some of our 'stable' features start deviating from main purpose and not be relevant to most users

In that case we can allow the specific users to enable / disable those features

src/Angor/Client/Services/FeatureFlagService.cs
theDRB123: Fixed

src/Angor/Client/Storage/ClientStorage.cs
```diff
@@ -223,5 +223,16 @@ public void DeleteSignatures()
...
+    public void setFeatureFlags(Dictionary<string, bool> featureFlags)
+    {
+        _storage.SetItem("FeatureFlags", featureFlags); 
+    }
+
+    public Dictionary<string, bool> getFeatureFlags()
+    {
+        return _storage.GetItem<Dictionary<string, bool>>("FeatureFlags") ?? new()
+        {
+            {"HW_Support", false}
```
dangershony, line 235: hmm I am not a fan of putting our default feature flags in the storage class, it should come form the config file `INetworkConfiguration` 
But I am not sure how we can pass it to here?
Perhaps inject the interface and return the feature flag collection from here?

src/Angor/Client/Storage/IClientStorage.cs
```diff
@@ -29,5 +29,8 @@ public interface IClientStorage
 
         string GetCurrencyDisplaySetting();
         void SetCurrencyDisplaySetting(string setting);
+
+        void setFeatureFlags(Dictionary<string, bool> featureFlags);
```
dangershony, line 33: yes this is what I was looking for

dangershony (commented):
Almost ready one more small change to more the flags to config

src/Angor/Client/Storage/ClientStorage.cs
theDRB123: yeah that will be much better, on it

src/Angor/Client/Pages/Settings.razor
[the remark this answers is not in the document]
theDRB123: If we really want the feature flags for both the mainnet and testnet, I think what we can do that while injecting the NetworkConfiguration, and then in it the method would return the dictionary of features depending on the network selected(current network)...

src/Angor/Client/Storage/ClientStorage.cs
theDRB123: Implemented in the latest commit

src/Angor/Client/Pages/Settings.razor
```diff
@@ -591,9 +614,15 @@
...
+        _featureFlags = _featureFlagService.GetAllFeatureFlags();
+
         return base.OnInitializedAsync();
     }
-
+    private void OnFeatureFlagChanged(string key, bool value)
+    {
+        _featureFlagService.SetFeatureFlag(key, value);
+        _featureFlags[key] = value;
+    }
```
dangershony, line 625: leave space between methods (fix it in a future pr)

dangershony (approved):

PR #373, opened 2025-05-13
Sanitize the content received from the relay

dangershony:


src/Angor/Client/Services/MessageService.cs
```diff
@@ -15,6 +16,7 @@ public class MessageService : IMessageService
         private readonly INetworkService _networkService;
         private readonly IRelayService _relayService;
         private readonly NostrConversionHelper _nostrHelper;
+        private readonly HtmlSanitizer _sanitizer;
```
DavidGershony, line 19: I think this should be wrapped in an interface so we can replace it if we want to

dangershony: yeah I agree

PR #375, opened 2025-05-14
Check if founder has pending messages

dangershony:


src/Angor/Client/Pages/Signatures.razor
```diff
@@ -589,6 +599,27 @@
             });
     }
 
+    Dictionary<string, string> pubkeyWithMessages = new();
+
+    protected async Task ScanForDmMessages()
+    {
+        foreach (var signatureRequest in signaturesRequests)
```
DavidGershony, line 606: There is a message type called count in nostr that you send the same filter and it returns how many events are there for it

dangershony:
This is still an experimental PR so dont revie wit yet

src/Angor/Client/Pages/Signatures.razor
DavidGershony: Good question, perhaps we can implement something based on support for each relay in the nostr-client

PR #378, opened 2025-05-14
Update PasswordComponent to use radio buttons

miladsoft:


src/Angor/Client/Shared/PasswordComponent.razor
```diff
@@ -90,10 +103,26 @@
...
 
     public Func<Task> Submit { get; set; }
 
+    private void HandleCacheSelection(string value)
+    {
+        SelectedRadioValue = value;
+        if (value == "active")
+        {
+            _passwordCacheService.UnlockWhileActive = true;
+            SelectedCacheDuration = 1;
```
dangershony, line 117: I wonder if we should default this to 5min?

dangershony (approved):

PR #381, opened 2025-05-15
UI and Service Fix for featureflags

theDRB123:
Changes:
- Added a setallfeatureflags method
- Feature flags now don't change directly on ticking the box, instead user needs to click on "Change Network Settings"
- This would make wallet reset mandatory when changing the flags
image

### DIFF
--- a/src/Angor/Client/Pages/Settings.razor
+++ b/src/Angor/Client/Pages/Settings.razor
@@ -554,7 +554,6 @@
     }
     private void OnFeatureFlagChanged(string key, bool value)
     {
-        _featureFlagService.SetFeatureFlag(key, value);
         _featureFlags[key] = value;
     }
     private void OnNetworkChange(string network){
@@ -911,10 +910,6 @@
             notificationComponent.ShowErrorMessage("Failed to wipe data: " + e.Message);
         }
     }
-    private void ChangeFeatureFlags()
-    {
-        _featureFlagService.SetAllFeatureFlags(_featureFlags);
-    }
 
     private void ChangeNetwork()
     {
@@ -939,7 +934,7 @@
             _networkService.CheckAndSetNetwork(_navManager.Uri.ToLower(), selectedNetwork);
             _networkService.AddSettingsIfNotExist();
 
-            ChangeFeatureFlags();
+            _featureFlagService.SetAllFeatureFlags(_featureFlags);
 
             // Refresh UI data
             networkType = _networkConfiguration.GetNetwork().Name;
--- a/src/Angor/Client/Pages/Settings.razor
+++ b/src/Angor/Client/Pages/Settings.razor
@@ -105,11 +105,12 @@
                                             <div class="info-card mb-3">
                                                 <label for="networkSelection" class="info-label m-0">Select Network:</label>
                                                 <select class="form-control mt-2" id="networkSelection"
-                                                        @bind="selectedNetwork">
-                                                    <option value="Angornet" selected>Angornet</option>
-                                                    <option value="TestNet4" disabled selected>Testnet4</option>
-                                                    <option value="Signet" disabled selected>Signet</option>
-                                                    <option value="TestNet" disabled selected>Testnet</option>
+                                                        value="@selectedNetwork"
+                                                        @onchange="(e) => OnNetworkChange(e.Value?.ToString())">
+                                                    <option value="Angornet">Angornet</option>
+                                                    <option value="TestNet4" disabled>Testnet4</option>
+                                                    <option value="Signet" disabled>Signet</option>
+                                                    <option value="TestNet" disabled>Testnet</option>
                                                     <option value="Main">Mainnet</option>
                                                     <option value="Liquid" disabled>Liquid</option>
                                                 </select>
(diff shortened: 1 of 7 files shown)


dangershony:
> This would make wallet reset mandatory when changing the flags

Ahh I didn't notice that was not the case, good spot.

src/Angor/Client/Pages/Settings.razor
```diff
@@ -906,6 +906,20 @@
             notificationComponent.ShowErrorMessage("Failed to wipe data: " + e.Message);
         }
     }
+    private void ChangeFeatureFlags()
+    {
+        var newFlags = _featureFlagService.GetAllFeatureFlags();
```
dangershony, line 911: isn't this name misleading? it should be `currentFlags`

src/Angor/Client/Pages/Settings.razor
```diff
@@ -906,6 +906,20 @@
...
+
+        foreach (var (key, value) in newFlags)
+        {
+            if (_featureFlags.TryGetValue(key, out var currentValue))
+            {
+                newFlags[key] = currentValue;
+            }
+        }
+        _featureFlags = newFlags;
```
dangershony, line 920: I wonder why do all this and not just push the new set flags?

theDRB123: Basically what I have done here is, changing either feature flags or network with the same button,
So if two networks have the same feature flags then preserve the feature flag settings between them, otherwise remove feature flag if not present in the new network..

Should I change it to have two buttons for feature flag and network ? Might be a better ux

src/Angor/Client/Pages/Settings.razor
theDRB123: Yeah i will change that

src/Angor/Client/Pages/Settings.razor
dangershony: I don't think we need to preserve past selections, when you change network wipe all feature flag data and set the new data that is selected, the whole idea of wiping the data is clean all the storage so why not clean it also for feature flags?

src/Angor/Client/Pages/Settings.razor
```diff
@@ -906,6 +911,10 @@
             notificationComponent.ShowErrorMessage("Failed to wipe data: " + e.Message);
         }
     }
+    private void ChangeFeatureFlags()
```
dangershony, line 914: if this method is called only once then I rather delete it and put the code in the `ChangeNetwork` method

src/Angor/Client/Services/FeatureFlagService.cs
```diff
@@ -1,15 +1,18 @@
...
-    private readonly Dictionary<string, bool> _featureFlags;
-    public FeatureFlagService(IClientStorage storage)
+    private readonly INetworkConfiguration _networkConfiguration;
+    private Dictionary<string, bool> _featureFlags;
+    public FeatureFlagService(IClientStorage storage, INetworkConfiguration networkConfiguration)
     {
         _storage = storage;
-        _featureFlags = _storage.getFeatureFlags();
+        _networkConfiguration = networkConfiguration;
+        _featureFlags = _storage.getFeatureFlags() ?? [];
```
dangershony, line 15: As a rule try to not do anything heavy in contractors, they are called by the service dependency injector and should be as light weight as possible.
you can move this code to the `GetAllFeatureFlags()` method like `_featureFlags ?? _storage.getFeatureFlags()`

src/Angor/Client/Services/FeatureFlagService.cs
```diff
@@ -28,7 +31,17 @@ public void SetFeatureFlag(string featureName, bool isEnabled)
 
     public Dictionary<string, bool> GetAllFeatureFlags()
     {
-        return _storage.getFeatureFlags();
+        return _featureFlags;
+    }
+    public Dictionary<string, bool> GetDefaultFeatureFlags(string network)
```
dangershony, line 36: spaces between methods

src/Angor/Client/Services/IFeatureFlagService.cs
```diff
@@ -8,5 +6,7 @@ public interface IFeatureFlagService
     public bool IsFeatureEnabled(string featureName);
     public void SetFeatureFlag(string featureName, bool isEnabled);
     public Dictionary<string, bool> GetAllFeatureFlags();
+    public void SetAllFeatureFlags(Dictionary<string, bool> featureFlags);
```
dangershony, line 9: if we set all flags in one go do we still need the method `SetFeatureFlag`

src/Angor/Client/Storage/IClientStorage.cs
```diff
@@ -31,6 +31,6 @@ public interface IClientStorage
         void SetCurrencyDisplaySetting(string setting);
 
         void setFeatureFlags(Dictionary<string, bool> featureFlags);
-        Dictionary<string, bool> getFeatureFlags();
+        Dictionary<string, bool>? getFeatureFlags();
```
dangershony, line 34: `setFeatureFlags => SetFeatureFlags`
`getFeatureFlags => GetFeatureFlags`

dangershony (approved):

PR #388, opened 2025-05-16
Add Avalonia feature: founder investment approval

SuperJMN:
Founders can approve investment requests.


SuperJMN:
Batch call still pending

src/Angor/Avalonia/Angor.Contexts.CrossCutting/NostrQueryClient.cs
```diff
@@ -0,0 +1,75 @@
...
+
+namespace Angor.Contests.CrossCutting;
+
+/// <summary>
+/// Client that provides query capabilities for Nostr events through multiple WebSocket connections.
+/// Completes each per-relay event stream upon its EOSE and merges them.
+/// </summary>
+public class NostrQueryClient
+{
+    private readonly INostrClient[] clients;
```
DavidGershony, line 14: The NuGet package supports a single client that connects to all relays, do you think it's better to manually do it? I think they basically forward every event from each connection handler to the client streams

src/Angor/Avalonia/Angor.Contexts.Funding/Investor/IInvestmentAppService.cs
```diff
@@ -8,8 +8,8 @@ namespace Angor.Contexts.Funding.Investor;
 
 public interface IInvestmentAppService
```
DavidGershony, line 9: Jose, perhaps the app service interfaces should be the only things in the project that are public, the rest can be internal for the testing project to use

src/Angor/Avalonia/Angor.Contexts.Funding/Shared/INostrDecrypter.cs
```diff
@@ -1,10 +1,11 @@
...
 using Angor.Contexts.Funding.Projects.Domain;
+using Angor.Contexts.Funding.Shared;
 using CSharpFunctionalExtensions;
 
 namespace Angor.Contexts.Funding;
 
 public interface INostrDecrypter
 {
-    Task<Result<string>> Decrypt(Guid walletId, ProjectId projectId, NostrMessage nostrMessage);
+    Task<Result<string>> Decrypt(Guid walletId, ProjectId projectId, InvestmentMessage nostrMessage);
```
DavidGershony, line 10: This should be DM message

src/Angor/Avalonia/Angor.Contexts.Funding/FundingContextServices.cs
```diff
@@ -56,8 +59,26 @@ public static ServiceCollection Register(ServiceCollection services, ILogger log
...
         services.TryAddSingleton<IWalletOperations, WalletOperations>();
+        services.AddSingleton(GetNostrQueryClient);
         services.AddHttpClient();
 
         return services;
     }
+
+    private static NostrQueryClient GetNostrQueryClient(IServiceProvider provider)
+    {
+        var nc = provider.GetRequiredService<INetworkConfiguration>();
```
DavidGershony, line 70: I think this kind of business logic should be in a service, it's too explicit and related to the internals of a NuGet package we consume for nostr-client

DavidGershony (changes requested):
I think the nostr client code in the service provider binding should move otherwise it's all good

src/Angor/Avalonia/Angor.Contexts.CrossCutting/NostrQueryClient.cs
```diff
@@ -0,0 +1,75 @@
...
+        this.clients = clients;
+    }
+
+    /// <summary>
+    /// Queries events matching the filter. Completes when each client has signaled EOSE or the timeout elapses.
+    /// </summary>
+    /// <param name="filter">Filter to apply to events.</param>
+    /// <param name="timeout">Maximum time to wait for events.</param>
+    /// <returns>An observable sequence of <see cref="NostrEventResponse"/>.</returns>
+    public IObservable<NostrEventResponse> Query(NostrFilter filter, TimeSpan timeout)
```
dangershony, line 34: this is interesting, so it is a general query class for nostr, it may also need to take count not just date, how will the subscription be returned to the caller in case they want to close the connection later?

src/Angor/Avalonia/Angor.Contexts.CrossCutting/NostrQueryClient.cs
SuperJMN: Do you mean the `NostrMultiWebsocketClient`? I actually started with that approach, but I ultimately decided on injecting an array of INostrClient instances for three reasons:

**Loose coupling:** by coding against the interface, the query client remains agnostic to any specific implementation. Changes in a concrete multi‐websocket class won’t inadvertently break its behavior.

**Immutable configuration:** avoiding a mutable client list removes the risk of race conditions when multiple queries run concurrently. Each invocation works from its own fixed set of endpoints.

**Single responsibility:** separating relay management from query orchestration keeps each class focused and more easily tested.

That said, I’m relatively new to Nostr and could easily have overlooked a better abstraction. If you see a flaw in this design or a compelling reason to revisit the decorator pattern, please tell me :)

src/Angor/Avalonia/Angor.Contexts.Funding/Investor/IInvestmentAppService.cs
SuperJMN: You're absolutely right! Let's do that in a separate PS. I'll create a card so I don't forget it.

src/Angor/Avalonia/Angor.Contexts.Funding/Shared/INostrDecrypter.cs
SuperJMN: Done!

src/Angor/Avalonia/Angor.Contexts.CrossCutting/NostrQueryClient.cs
SuperJMN: Thanks! I return an `IObservable<NostrEventResponse>` so that callers can control count, timeout and disposal themselves:

```csharp
// Example: limit to 10 events and 5‐second timeout
var subscription = client.Query(filter)
                         .Take(10)                   // maximum 10 events
                         .Timeout(TimeSpan.FromSeconds(5))
                         .Subscribe(response => { /* handle event */ });

// When you call subscription.Dispose(), it unsubscribes
// and triggers our Finally() block to send NostrCloseRequest.
```

**Count control:** .Take(maxCount) enforces any desired limit.

**Subscription handle:** .Subscribe(...) returns an IDisposable. Disposing it both unsubscribes and closes the relay streams.

**Flexible composition:** You can layer in buffering, retry, merging, etc., without modifying the core API.

If you need a pull‐style list, you can still write:

```csharp
var list = await client.Query(filter)
                       .Take(maxCount)
                       .ToList()
                       .Timeout(timeout)
                       .ToTask();

```

That produces an `IReadOnlyList<NostrEventResponse>` with no changes to Query. Let me know if that addresses your scenario or if a higher‐level helper method would simplify your usage.

I think this covers a lot of use cases in our business logic. I'd like to know your opinion on it :) We can use it and see how it goes. We can even create a decorator over it that transparently decrypts the events as discussed with @DavidGershony.

src/Angor/Avalonia/Angor.Contexts.Funding/FundingContextServices.cs
SuperJMN: Yes, I agree. Anyways, I've removed the registration completely for now because, in the end, the client isn't used at all. We might even remove it completely from this PR and add it in more scoped PR. I'm good with both. What do you prefer?

SuperJMN:
Is this missing something @DavidGershony ?

dangershony (approved):

PR #392, opened 2025-05-20
Prepare method to scan for how an investment was spent when recovering

dangershony:


src/Angor/Client/Pages/Investor.razor
```diff
@@ -631,17 +635,104 @@
...
+                        {
+                            resolved = true;
+                            investorProject.UnfundedReleaseTransactionId = output.SpentInTransaction;
+                            break;
+                        }
+                    }
+                }
+            }
+
+            if (resolved)
```
DavidGershony, line 728: This boolean is really confusing, is there no other way?

DavidGershony (commented):
I think we shouldn't have setting of booleans in the loop, better to just call the method in each case or just have default with false and only than not call it.

PR #407, opened 2025-05-31
Improved the structure and flow of the project creation wizard

miladsoft:


src/Angor/Client/Pages/Create.razor
```diff
@@ -492,207 +497,270 @@
...
+                        </button>
+                    </div>
+                </div>
+            }
+            else if (hasErrors)
+            {
+                <div class="text-center mb-4">
+                    <div class="d-inline-flex align-items-center p-2 rounded border border-warning">
+                        <Icon IconName="warning" Width="16" Height="16" Class="me-2 text-warning" />
+                        <span class="fw-semibold text-warning">Manual Mode Required</span>
```
dangershony, line 534: I dont think this is what I had in mind, basically we just need one button that say continue and if we failed we do the prev step again right? so it should always be automatic mode

dangershony (dismissed):
looks ok otherwise

src/Angor/Client/Pages/Create.razor
```diff
@@ -489,210 +494,309 @@
...
-                    <div class="d-flex flex-column align-items-center position-relative z-1">
-                        <div class="step-circle d-flex align-items-center justify-content-center rounded-circle @(currentStep >= 1 ? "bg-success text-white" : " border")"
-                             style="width: 40px; height: 40px; transition: all 0.3s ease;">
-                            @if (currentStep > 1)
+            @if (hasErrors)
+            {
+                <div class="text-center mb-4">
+                    <div class="alert alert-danger d-flex align-items-center justify-content-center">
+                        <Icon IconName="warning" Width="20" Height="20" Class="me-2 text-warning" />
+                        <span class="fw-semibold">Deployment Failed - try again</span>
```
dangershony, line 510:
```suggestion
                        <span class="fw-semibold">Deployment Failed - you can try again</span>
```

dangershony (approved):

PR #413, opened 2025-06-06
Indicator for Unseen messages on Approve Investment button

yashop7:
Indicator appears on the button if there are pending messages. After reading them, the indicator goes away

Screenshot 2025-06-07 at 5 10 38 AM

https://github.com/user-attachments/assets/98442778-f68a-4779-a199-fc0cf77257cf

### DIFF
--- a/src/Angor/Client/Pages/View.razor
+++ b/src/Angor/Client/Pages/View.razor
@@ -878,7 +878,6 @@
             {
                 founder = true;
                 NavMenuState.SetActivePage("founder");
-                // If project belongs to founder, fetch pending signatures
                 if (project is FounderProject founderProject)
                 {
                     await _RelayService.LookupSignaturesDirectMessagesForPubKeyAsync(founderProject.ProjectInfo.NostrPubKey, founderProject.LastRequestForSignaturesTime?.AddSeconds(1), 1,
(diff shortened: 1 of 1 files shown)


src/Angor/Client/Pages/View.razor
```diff
@@ -960,6 +976,73 @@
         }
     }
 
+    private bool hasPendingMessages = false;
+
+    HashSet<string> pubkeyWithMessages = new();
+
+    private async Task FetchPendingSignatures(FounderProject founderProject)
```
dangershony, line 983: I don't understand this code really, we just need to find out if there are pending messages 

All you need to do here is call the method
```
        await RelayService.LookupSignaturesDirectMessagesForPubKeyAsync(FounderProject.ProjectInfo.NostrPubKey, FounderProject.LastRequestForSignaturesTime?.AddSeconds(1), 1,
            _ =>
            {
                if (InvestmentRequests)
                    return;

                InvestmentRequests = true;
                StateHasChanged();
            });
```

and if there are InvestmentRequests  then mark the button

src/Angor/Client/Pages/View.razor
```diff
@@ -1291,5 +1374,21 @@
             StateHasChanged();
         }
     }
+
+    public class SignatureItem
```
dangershony, line 1378: delete this no need for this class here, delete all code that is not part of the method I mentioned

src/Angor/Client/Pages/View.razor
```diff
@@ -869,6 +878,18 @@
             {
                 founder = true;
                 NavMenuState.SetActivePage("founder");
+                if (project is FounderProject founderProject)
+                {
+                    await _RelayService.LookupSignaturesDirectMessagesForPubKeyAsync(founderProject.ProjectInfo.NostrPubKey, founderProject.LastRequestForSignaturesTime?.AddSeconds(1), 1,
```
DavidGershony, line 883: Is there actually a need to add a second to the time?

DavidGershony (commented):
Need to check the NIP

dangershony (approved):

dangershony:
> Need to check the NIP

indeed but it is the same as in the sigs page so its fair to say it works fine (worked till now)

PR #415, opened 2025-06-09
Stop Background scroll behind Modal

yashop7:
- wrote a Service which will insert & remove the (.modal-open) class in the body Tag
- We have to call it whenever making state T/F for calling the modal


yashop7:
Hey @dangershony, can you please check the implementation? Once you green light it, I will propagate this to other modals in the code

PR #416, opened 2025-06-09
Improve backend API, AKA "Brainstorming with David"

SuperJMN:
Includes all the changes that @DavidGershony and @SuperJMN to improve abstractions. Encryption is handled by a deeper layer instead of having the caller encrypt everything. 

Notice that only a few methods work like this as part of the PR.

### DIFF
--- a/src/Angor/Avalonia/Angor.Contexts.CrossCutting/NetworkStorage.cs
+++ b/src/Angor/Avalonia/Angor.Contexts.CrossCutting/NetworkStorage.cs
@@ -1,7 +1,7 @@
 using Angor.Shared;
 using Angor.Shared.Models;
 
-namespace Angor.Contexts.Wallet.Infrastructure.Impl;
+namespace Angor.Contests.CrossCutting;
 
 public class NetworkStorage : INetworkStorage
 {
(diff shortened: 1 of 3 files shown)


SuperJMN:
> I let @DavidGershony do avalonia PRs but you have conflicts

Just found out that a previously merged PR caused conflicts. I've just solved them :)

DavidGershony (approved):

src/Angor/Client/NetworkConfiguration.cs
```diff
@@ -63,8 +63,8 @@ public List<SettingsUrl> GetDefaultIndexerUrls()
             {
                 return new List<SettingsUrl>
                 {
-                    new SettingsUrl { Name = "", Url = "https://test.indexer.angor.io", IsPrimary = true },
                     new SettingsUrl { Name = "", Url = "https://mempool.thedude.pro", IsPrimary = false },
+                    new SettingsUrl { Name = "", Url = "https://test.indexer.angor.io", IsPrimary = false },
```
dangershony, line 67: there is no primary here

src/Angor/Shared/Angor.Shared.csproj
```diff
@@ -11,11 +11,12 @@
 
   <ItemGroup>
     <PackageReference Include="Blockcore.Core" Version="1.1.41" />
-    <PackageReference Include="Microsoft.Extensions.Http" Version="9.0.5" />
+    <PackageReference Include="CSharpFunctionalExtensions" Version="3.6.0" />
```
dangershony, line 14: what is this change?

PR #421, opened 2025-06-11
Seed Phrase Suggestion Bug

yashop7:
- Suggestions gone after you removed focus from the particular Tab
- Suggestion also gone after you fully typed the phrase
- Minor change in Placeholder text 

Before:

https://github.com/user-attachments/assets/d4a8d1e1-69f3-4b3f-bd45-662617f12670

After:

https://github.com/user-attachments/assets/e96a650a-7bb0-49a8-8184-c09c532b6faf

### DIFF
--- a/src/Angor/Client/Pages/Wallet.razor
+++ b/src/Angor/Client/Pages/Wallet.razor
@@ -163,12 +163,13 @@
                                                         var wordIndex = selectedWordsToVerify[i].index;
                                                         <div class="col-md-4 col-sm-6 mt-0">
                                                             <div class="mb-3 position-relative">
-                                                                <label class="form-label">Word #@wordIndex</label>
+                                                                <label class="form-label">Word @wordIndex</label>
                                                                 <input type="text" class="form-control"
                                                                        value="@userVerificationInputs[wordIndex]"
                                                                        @oninput="(e) => HandleWordInput(wordIndex, e)"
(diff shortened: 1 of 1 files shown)


src/Angor/Client/Pages/Wallet.razor
```diff
@@ -178,7 +179,7 @@
                                                                !string.IsNullOrWhiteSpace(userVerificationInputs[wordIndex]))
                                                                 {
                                                                     var suggestions = GetFilteredWordSuggestions(userVerificationInputs[wordIndex]);
-                                                                    @if (suggestions.Any())
+                                                                    if (!(suggestions.Count() == 1 && suggestions.First().Equals(userVerificationInputs[wordIndex], StringComparison.OrdinalIgnoreCase)) && suggestions.Any())
```
dangershony, line 182: what is the point of this line?

yashop7: On it now!

yashop7: Yeah Dan, you are right, made the change

yashop7:
the issue was that onblur get's triggered before onclick, that's why when we click some option on the dropdown, in meantime onblur removes the focus so option selected wasn't reflected, so now I added a delay in onblur, now it's working fine, In my case earlier onclick was triggering before onblur, so it was working fine for me

dangershony (approved):

PR #423, opened 2025-06-12
Migrate to .NET 9 and configure HttpClient with BaseAddress

miladsoft:


src/Angor/Client/Program.cs
```diff
@@ -20,6 +20,10 @@
 
 builder.Services.AddHttpClient();
 
+builder.Services.AddScoped(sp => new HttpClient
```
dangershony, line 23: why is this needed to fix doent9?

PR #424, opened 2025-06-12
Changed the dependency to the Microsoft NuGet package

DavidGershony:


src/Angor/Shared/Angor.Shared.csproj
```diff
@@ -12,11 +12,11 @@
   <ItemGroup>
     <PackageReference Include="Blockcore.Core" Version="1.1.41" />
     <PackageReference Include="CSharpFunctionalExtensions" Version="3.6.0" />
+    <PackageReference Include="Microsoft.Extensions.Http" Version="9.0.6" />
```
dangershony, line 15: does this fix the dotnet9 bug?

PR #425, opened 2025-06-12
Fix progress bar width calculation in FounderProjectItem component

miladsoft:


dangershony:
@miladsoft you will have to cherry pick this pr to current master sorry

PR #429, opened 2025-06-15
Upgrade dotnet9

dangershony:


dangershony:
FYI @SuperJMN and @DavidGershony 
I have tested this on linux popos box running in localhost and it is working fine.
So the issue iis really the deployment to github pages

dangershony:
Closing this PR as it will never go in but we can keep the branch for now

PR #434, opened 2025-06-18
Improve UI

SuperJMN:
- Adds Dark mode
- More Responsive


src/Angor/Avalonia/Angor.UI.Model/IProject.cs
```diff
@@ -4,8 +4,8 @@ public interface IProject
 {
     public string Id { get; set; }
     public string Name { get; }
-    public Uri Picture { get; }
     public Uri Banner { get; }
+    public Uri Picture { get; }
```
DavidGershony, line 8: We should have a todo to set it to nullable, we might not get a banner and picture.

PR #445, opened 2025-06-26
Adding Cached Indexer Service

theDRB123:
Added cached indexer service to reduce indexer calls and improve responsiveness

modifications:

- Added a new IIndexerService implementation with cached indexer calls, which uses MempoolSpaceIndexerApi service directly 
- Removed binding of MempoolSpaceIndexerApi from the builder
- Modified dependency on shared and client

Edit:
- Instead of IIndexerService implementation, a ICachedIndexer interface is introduced, for further extension & to be used only where required 

Problems to be resolved:
- A parameter for disabling caching is required wherever updated data is necessary (refresh wallet etc) [Fixed]

### DIFF
--- a/src/Angor/Shared/Services/CachedIndexerService.cs
+++ b/src/Angor/Shared/Services/CachedIndexerService.cs
@@ -107,7 +107,7 @@ namespace Angor.Shared.Services
             });
         }
 
-        public async Task<ProjectStats?> GetProjectStatsAsync(string projectId)
+        public async Task<(string projectId, ProjectStats? stats)>GetProjectStatsAsync(string projectId)
         {
             var cacheKey = $"project_stats_{projectId}";
             return await _cache.GetOrCreateAsync(cacheKey, async entry =>


src/Angor/Shared/Services/CachedIndexerService.cs
```diff
@@ -0,0 +1,152 @@
...
+
+namespace Angor.Shared.Services
+{
+    public class CachedIndexerService : ICachedIndexerService
+    {
+        private readonly IIndexerService _indexer;
+        private readonly IMemoryCache _cache;
+        private readonly TimeSpan _cacheExpiration = TimeSpan.FromMinutes(5);
+
+        public CachedIndexerService(IIndexerService indexerService, IMemoryCache cache)
```
DavidGershony, line 12: We don't need to cache every method, just the ones that you identify as problematic, for the rest of them just call directly to the MempoolSpaceIndexerApi you have as a property

src/Angor/Client/Pages/Spend.razor
```diff
@@ -766,89 +766,104 @@
...
-                }
+    private async Task<Transaction> GetTransactionCache(string txid)
+    {
+        if (transactionCache.TryGetValue(txid, out var cachedTask))
+            return await cachedTask;
 
-                var output = trx.Outputs.First(outp => outp.Index == item.Outputindex);
+        var task = Task.Run(async () =>
+        {
+            var fetchTrx = await _IndexerService.GetTransactionHexByIdAsync(txid);
```
DavidGershony, line 808: Better to only return the task from the indexer service and call the network create transaction somewhere else

PR #447, opened 2025-06-27
Improve Portfolio section (Avalonia)

SuperJMN:
image

image


src/Angor/Avalonia/Angor.Contexts.Funding/Investor/InvestedProjectDto.cs
```diff
@@ -12,7 +13,7 @@ public class InvestedProjectDto
     public Amount Raised { get; set; }
     public string Description { get; set; }
     public Amount InRecovery { get; set; }
-    public bool IsInvesmentCompleted { get; set; }
+    public InvestmentStatus InvestmentStatus { get; set; }
```
DavidGershony, line 16: Do we still need the Founder status in this case?

SuperJMN: I don't know if it's relevant for the investor. Shall we remove it from the invested projects?

PR #448, opened 2025-06-28
Nip65 relay list

dangershony:


src/Angor/Shared/Services/RelayService.cs
```diff
@@ -276,8 +278,30 @@ public Task<string> DeleteProjectAsync(string eventId, string hexPrivateKey)
             
             return Task.FromResult(deleteEvent.Id);
         }
-        
-        private static NostrEvent GetNip3030NostrEvent( string content)
+
+        public Task<string> CreateNostrRelayNip65Async(string hexPrivateKey)
+        {
+            Nostr.Client.Messages.Contacts.NostrRelays relayList = new NostrRelays(_networkService.GetRelays()
```
DavidGershony, line 284: Do we actually need to use the NostrRelays and NostrRelayConfig here? as far as I can tell we just use the names.

DavidGershony (dismissed):
Added a single comment

PR #462, opened 2025-07-10
Mainnet Rule Mode Toggle UI

mistic0xb:
-  True: mainnet cases are applied 
-  False: mainnet cases are skipped (normal testnet)


src/Angor/Client/Pages/Settings.razor
```diff
@@ -126,9 +126,7 @@
                                                     @foreach (var feature in _featureFlags)
                                                     {
                                                         <div class="form-check mb-3">
-                                                            <input class="form-check-input"
-                                                                type="checkbox"
-                                                                id="@feature.Key"
+                                                            <input class="form-check-input" type="checkbox" id="@feature.Key"
```
dangershony, line 129: you have changes that are unrelated, no need to make such changes, the only changes in the ui should be the additional section

src/Angor/Client/Pages/Settings.razor
```diff
@@ -393,6 +393,44 @@
 
 </div>
 
+@* Mainnet_Rules_Mode Toggle
```
dangershony, line 396: remove this explanation it just clutters the ui

mistic0xb: okay

PR #468, opened 2025-07-14
Use PSBT for fee calculation

SuperJMN:
Fixes fee calculation using PSBT. It got broken due to changes in the **IWalletOperations** interface.


src/Angor/Avalonia/Angor.Contexts.Wallet/Infrastructure/Impl/WalletAppService.cs
```diff
@@ -182,4 +198,29 @@ public string GenerateRandomSeedwords()
         var mnemonic = new Mnemonic(Wordlist.English, WordCount.Twelve);
         return mnemonic.ToString();
     }
+
+    private async Task<Result<long>> CalculateTransactionFee(SendInfo sendInfo, AccountInfo accountInfo, WalletWords walletWords, DomainFeeRate feeRate)
```
dangershony, line 202: you are only doing this for the fee size? I suggest keeping the signedTransaction no? so that you wont need to build it again (not that it matter on desktop is it probably pretty fast)

dangershony: There is a property on sendInfo you could just store the trx there

dangershony (approved):
Looks ok to me, my comment are more about design but that is not related to this pr so much

src/Angor/Avalonia/Angor.Contexts.Wallet/Infrastructure/Impl/WalletAppService.cs
dangershony: I don't think caching will help here, and In any case to get real fee estimation you need to create the signed trx and know its exact size, so it is hard to get the fee only (accurately) without building the entire signed trx.
If this is a perf issue we can revist

src/Angor/Avalonia/Angor.Contexts.Wallet/Infrastructure/Impl/WalletAppService.cs
```diff
@@ -182,4 +198,29 @@ public string GenerateRandomSeedwords()
...
+            var psbt = psbtOperations.CreatePsbtForTransaction(
+                unsignedTransaction, 
+                accountInfo, 
+                feeRate.SatsPerVByte, 
+                utxoDataWithPaths: sendInfo.SendUtxos.Values.ToList()
+            );
+            
+            var signedTransaction = psbtOperations.SignPsbt(psbt, walletWords);
+            var realFeeInSatoshis = signedTransaction.TransactionFee;
+            var virtualSize = (long)signedTransaction.Transaction.GetVirtualSize(0);
```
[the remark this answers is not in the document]
dangershony: it should be `.GetVirtualSize(4)`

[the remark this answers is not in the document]
SuperJMN: Fixed

PR #475, opened 2025-07-16
Fix create project

miladsoft:


src/Angor/Client/Pages/Create.razor
```diff
@@ -1616,106 +1690,206 @@
...
+            return false;
+        }
+
+        if (targetAmountInBTC < (decimal)0.01)
+        {
+            notificationComponent.ShowErrorMessage("Project target amount must be at least 0.01 BTC.");
+            return false;
+        }
+
+        if (project.ProjectInfo.StartDate < DateTime.Today)
```
dangershony, line 1851: is this a new check?

src/Angor/Client/Pages/Create.razor
```diff
@@ -1616,106 +1690,206 @@
...
+        {
+            notificationComponent.ShowErrorMessage("Project start date must be in the future.");
+            return false;
+        }
+
+        if (project.ProjectInfo.StartDate > DateTime.Today.AddYears(1))
```
dangershony, line 1857: is this a new check?

src/Angor/Client/Shared/MainLayout.razor
```diff
@@ -143,13 +158,36 @@
...
+    private string GetBaseNetworkName()
+    {
+        var network = _networkConfiguration.GetNetwork();
+        return network.Name == "Angornet" ? "Angornet" : network.Name;
+    }
+
+    private bool IsTestnet()
+    {
+        var network = _networkConfiguration.GetNetwork();
+        return network.Name == "Angornet";
```
dangershony, line 185: this should be checking the opposite `network.Type != mainnet`

dangershony (approved):
This is good now, we still have an issue that lots of the validation logic is on the page itself we would need to move it out to the service `IApplicationLogicService` at some point so it can be used by avalonia

PR #481, opened 2025-07-24
Add some logs to investor scan

dangershony:


dangershony:
@miladsoft lets wait before we push this in case @mistic0xb needs more logs

PR #494, opened 2025-08-18
Review Founder/Project flow

SuperJMN:
- Provide backend placeholder methods for Founder/Claim operation. 
- Adapt the frontend better to align the with business logic and simplify everything involved in the claim flow.
- Fixed potential crashes in ProjectRepository.

### DIFF
--- a/src/Angor/Avalonia/AngorApp/Sections/Portfolio/PortfolioSectionView.axaml
+++ b/src/Angor/Avalonia/AngorApp/Sections/Portfolio/PortfolioSectionView.axaml
@@ -87,7 +87,7 @@
                     </Card.HeaderEndContent>
                 </Card>
 
-                <CardGrid DockPanel.Dock="Top">
+                <CardGrid DockPanel.Dock="Top" Classes="ShowEmptyContent" Empty.Content="Nothing to show">
                     <CardGrid.ItemsSource>
                         <collections:ArrayList>
                             <Card Theme="{StaticResource RemarkableSubheaderCard}" Header="Funded Projects" Subheader="{Binding FundedProjects}">


src/Angor/Avalonia/Angor.Contexts.Funding/Projects/Application/Dtos/ProjectExtensions.cs
```diff
@@ -1,25 +0,0 @@
-namespace Angor.Contexts.Funding.Projects.Application.Dtos;
-
-// TODO: Check if the logic of these methods is correct
-public static class ProjectExtensions
-{
-    public static long Raised(this ProjectDto project)
-    {
-        return project.Stages.Sum(stage => stage.Amount);
```
DavidGershony, line 8: I am not sure the project dto holds the data for actual raised transactions, only the amount the stage will get in total

src/Angor/Avalonia/Angor.Contexts.Funding/Projects/Infrastructure/Interfaces/IProjectAppService.cs
```diff
@@ -8,6 +8,7 @@ public interface IProjectAppService
 {
     Task<Result<IEnumerable<ProjectDto>>> Latest();
     Task<Maybe<ProjectDto>> FindById(ProjectId projectId);
+    public Task<Result<ProjectDto>> Get(ProjectId projectId);
```
DavidGershony, line 11: No need for public, also why do we have findById and Get? can't we use just one for the same logic?

src/Angor/Avalonia/AngorApp/Sections/Browse/Details/InvestmentDashboard.axaml
```diff
@@ -4,8 +4,7 @@
              xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
              xmlns:dt="clr-namespace:AngorApp.Sections.Browse.Details"
              xmlns:i="https://github.com/projektanker/icons.avalonia"
-             mc:Ignorable="d" d:DesignWidth="600"
-             MinWidth="300"
+             mc:Ignorable="d" d:DesignWidth="600"
```
DavidGershony, line 7: I think we need to set a max width because on a large screen it doesn't look very good when it's stretched out.

DavidGershony (commented):
Just a few comments and some static methods that I am not sure are doing the correct thing

src/Angor/Avalonia/Angor.Contexts.Funding/Projects/Application/Dtos/ProjectExtensions.cs
SuperJMN: That file is deleted. The new one is namespace Angor.UI.Model.FullProjectExtensions.

src/Angor/Avalonia/Angor.UI.Model/FullProjectExtensions.cs
SuperJMN, line 1: Please @DavidGershony check this. The other extensions no longer exists.

PR #497, opened 2025-08-25
Remove unfunded limit

dangershony:
How to test this:
invest the exact amount and notice the total raised is 1% lower then target the code should still allow the founder to spend funds and the popup that says return funds to the investor is hidden

### DIFF
--- a/src/Angor/Shared/Services/ApplicationLogicService.cs
+++ b/src/Angor/Shared/Services/ApplicationLogicService.cs
@@ -29,9 +29,15 @@ namespace Angor.Shared.Services
             return false;
         }
 
-        public bool IsProjectFunded(long target, long current, bool started)
+        public bool IsProjectFunded(long target, long current)
         {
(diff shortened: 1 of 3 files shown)


src/Angor/Shared/Services/ApplicationLogicService.cs
```diff
@@ -28,5 +28,19 @@ public bool IsInvestmentWindowOpen(ProjectInfo? project)
...
             return false;
         }
+
+        public bool IsProjectFunded(long target, long current)
+        {
+            var feePercentage = _networkConfiguration.GetAngorInvestFeePercentage;
+
+            // reduce the fee from the total amount;
+            long angorFee = (target * feePercentage) / 100;
+            target -= angorFee;
```
DavidGershony, line 38: Are you sure we should reduce the fee from the target?

DavidGershony (commented):
not sure about the logic

PR #502, opened 2025-09-18
Improve Responsiveness, Stats Display, and Modal Functionality

miladsoft:


src/Angor/Client/Pages/Invest.razor
```diff
@@ -405,12 +405,12 @@
                             <div class="stage-card animate-fade-up">
                                 <div class="stage-header">
                                     <h6 class="mb-0">Stage @stage.StageNumber</h6>
-                                    <span class="badge bg-primary">@stage.AmountBtc @network.CoinTicker</span>
+                                    <span class="badge bg-primary">@stage.AmountBtc.ToString("0.00000000") @network.CoinTicker</span>
```
dangershony, line 408: why this change? the amount was not displayed correctly?

dangershony: ah true when we add too many digits, but what if we only have 0.001 will it show it as 0.0010000?

dangershony (approved):

PR #505, opened 2025-09-19
Refactor stage percentage calculation

miladsoft:


dangershony:
Can we change this pr so that we only allow full integer not decimal?

PR #507, opened 2025-09-23
Revert fix stage percentage adjustment and fix the percentage-split problem

mistic0xb:
- ### fix: added _DivideStages_ logic to stagePercentage which splits the percentage into round nums eliminating floating point issues
  - **earlier**:
    - let's say we have 100% and numOfStages = 6
    - earlier it was splitting the stagePercentage = 100 / 6 => 16.666..%
    - due to floating point precision, each stage showed 16.67% (when we round it from 16.6666...), but 16.67 × 6 = 100.02% (not exactly 100%)
    - this created rounding errors where total percentages didn't add up to exactly 100%
   - **current refactor**:
     - now stagePercentage uses _DivideStages_ logic to split into round nums for eg: [16,16,17,17,17,17]
     - this ensures that the totalPercentage adds up to exactly 100%

- ### fix: corrected stageDuration which had issues due to floating point precision error for eg:
  - **earlier**:
    - let's say the totalDurationOfProject = 100 and numOfStages = 6
    - earlier it was spiting the stageDuration = 100 / 6 => 16.66666... days
    - but due to stageDuration being and int, the precision was lost and stageDuration just 16
    - this effective lost 4 days in this case
  - **current refactor**:
    - now stageDuration is using the same logic _DivideStages_ to split the stages in round nums for eg: [16,16,17,17,17,17]
    - this ensures the totalDuration adds up to the totalDurationOfProject entered by the founder

### DIFF
--- a/src/Angor/Client/Pages/Create.razor
+++ b/src/Angor/Client/Pages/Create.razor
@@ -1941,20 +1941,22 @@
 
         project.ProjectInfo.Stages.Clear();
 
-        var stageDuration = totalDuration / numberOfStages;
-        var stagePercentage = 100m / numberOfStages;
+        var stageDurations = DivideStages(totalDuration,numberOfStages);
+        var stagePercentage = DivideStages(100,numberOfStages);
+        var cumulativeDuration = 0;
 
         for (int i = 0; i < numberOfStages; i++)
         {
             var stage = new Stage
             {
-                AmountToRelease = stagePercentage,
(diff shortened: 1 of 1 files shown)


dangershony:
I think the fix is actually to just use uint (no decimals) we do not need to split a % to decimals.
But we have to be careful with backwards compatibility, so basically we cannot just change the type to int but in the input we must immediately convert every value to int before it goes in the textbox

PR #510, opened 2025-09-29
Add theme switching functionality

SuperJMN:


DavidGershony (commented):
What about the svg transformer?

DavidGershony (approved):

PR #672, opened 2026-02-24
secure storage windows

mistic0xb:
# Secure Wallet Storage Implementation (Windows)

## Overview
This PR implements passwordless wallet encryption using Windows DPAPI (Data Protection API) and AES-256-GCM authenticated encryption. Users no longer need to enter or remember encryption passwords when creating or importing wallets.

### Two-Layer Encryption Architecture
The implementation uses an envelope encryption pattern with two distinct layers:

1. **OS-Level Protection (DPAPI)**
   - A unique 256-bit master key is generated per wallet
   - The master key is encrypted by Windows DPAPI bound to the current user account
   - DPAPI-encrypted keys are stored in `%AppData%\Angor\secure\{walletId}.key`
   - Keys can only be decrypted by the same Windows user on the same machine

2. **Wallet Data Encryption (AES-GCM)**
   - The master key encrypts wallet seed words and metadata using AES-256-GCM
   - GCM mode provides authenticated encryption (AEAD) — detects tampering automatically
   - Encrypted wallet data is stored in the existing LiteDB database

### Wallet Creation Flow
1. User enters seed words -> no password prompt
2. System generates a random 256-bit master key
3. Master key is encrypted with DPAPI -> saved to `%AppData%\Angor\secure\{walletId}.key`
4. Master key encrypts wallet data with AES-GCM
5. Master key bytes are cleared from memory

### Wallet Access Flow
1. Application loads encrypted wallet
2. System reads DPAPI-encrypted key file
3. DPAPI decrypts the master key silently (no user interaction)
4. Master key decrypts wallet data with AES-GCM
5. Seed words returned to application

## Security Properties

**What's Protected:**
- Master keys are bound to Windows user + machine (DPAPI CurrentUser scope)
- Additional entropy per wallet prevents key reuse attacks
- AES-GCM authentication tags detect any tampering with encrypted wallet data
- Sensitive data cleared from memory after use
- Key files marked as hidden system files

## Platform Support
**Current:** Windows only (DPAPI)  
**Future:** Cross-platform support planned via platform-specific secure storage:
- macOS: Keychain
- Linux
- Android: Android Keystore
- iOS: iOS Keychain

The AES-GCM encryption layer remains identical across platforms — only the master key storage mechanism changes.

## Breaking Changes
- `IWalletFactory.CreateWallet` signature changed — removed `encryptionKey` parameter
- UI wizard flows updated — removed encryption password step
- Existing wallets encrypted with user passwords are **not migrated automatically** — users must re-import from seed words

## Why Two Layers?
This approach separates concerns:
- **DPAPI layer:** Where the secret lives (OS-managed, platform-specific)
- **AES-GCM layer:** What it protects (portable, platform-independent)

Benefits:
- Easy to swap DPAPI for Apple-Keychain/Android-Keystore/Linux when going cross-platform

## Testing
Manual testing performed on Windows 11. Verify:
- Wallet creation completes without password prompt
- Wallet loads on app restart without password
- `.key` files created in `%AppData%\Angor\secure\`
- Multiple wallets each get unique keys
- Import wallet flow works without password


src/Angor/Avalonia/Angor.Sdk/Wallet/Infrastructure/Impl/DpapiSecureStorage.cs
```diff
@@ -0,0 +1,347 @@
...
+        }
+        catch (Exception ex)
+        {
+            return Result.Failure<string>($"Error updating master key: {ex.Message}");
+        }
+    }
+
+    private static string GenerateMasterKey()
+    {
+        using var rng = RandomNumberGenerator.Create();
```
DavidGershony: I think you need to use the cryptographically secure random number generator here

src/Angor/Avalonia/Angor.Sdk/Wallet/Infrastructure/Interfaces/IWalletImporter.cs
```diff
@@ -5,5 +5,5 @@ namespace Angor.Sdk.Wallet.Infrastructure.Interfaces;
 
 public interface IWalletFactory
 {
-    Task<Result<Domain.Wallet>> CreateWallet(string name, string seedwords, Maybe<string> passphrase, string encryptionKey, BitcoinNetwork network);
+    Task<Result<Domain.Wallet>> CreateWallet(string name, string seedwords, Maybe<string> passphrase, BitcoinNetwork network);
```
DavidGershony: We should have both options here.

DavidGershony (changes requested):
Needs some changes

dangershony:
superseded by this https://github.com/block-core/angor/pull/804

PR #794, opened 2026-04-29
fix: paginated tx fetching and mempool tx inclusion

dangershony:
## Summary

- **Bug**: `MempoolIndexerAngorApi` only fetched the first page (25 txs) from the indexer API. With >25 txs on a project address, the funding tx fell off and the project disappeared from Find Projects.
- Now paginates all pages using `?after_txid=` parameter until an empty page is returned
- Includes unconfirmed (mempool) transactions to catch recently broadcast txs
- Adds `MaxAddressTransactions` safety limit (1000) to prevent runaway fetches
- Fixes pagination termination: no longer assumes `page.Count < 25` means last page (some indexers return fewer items even when more pages exist)

## Changed files

- `src/shared/Angor.Shared/Services/MempoolIndexerAngorApi.cs`

Split from #789.

### DIFF
--- a/src/shared/Angor.Shared/Services/MempoolIndexerAngorApi.cs
+++ b/src/shared/Angor.Shared/Services/MempoolIndexerAngorApi.cs
@@ -42,6 +42,98 @@ public class MempoolIndexerAngorApi : IAngorIndexerService
         return client;
     }
 
+    /// <summary>
+    /// Fetches all transactions for an address by paginating through the mempool.space API.
+    /// The API returns max 25 confirmed txs per page (newest first). To get older pages,
+    /// pass ?after_txid={last_txid_from_previous_page}.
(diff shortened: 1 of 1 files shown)


src/shared/Angor.Shared/Services/MempoolIndexerAngorApi.cs
```diff
@@ -42,6 +42,98 @@ private HttpClient GetIndexerClient()
...
+
+        while (allTransactions.Count < MaxAddressTransactions)
+        {
+            var url = $"{MempoolApiRoute}/address/{projectAddress}/txs";
+            if (afterTxId != null)
+            {
+                url += $"?after_txid={afterTxId}";
+            }
+
+            var response = await GetIndexerClient().GetAsync(url);
```
DavidGershony: Should we get anew client for each call?

PR #804, opened 2026-04-30
feature/secure-key-provider

DavidGershony:
Added secure storage for the wallet encryption. also added ancrypting the cpubs in the database.


src/design/App/Composition/CompositionRoot.cs
```diff
@@ -107,11 +106,22 @@ public static IServiceProvider BuildServiceProvider(string profileName = "Defaul
...
+        // so TryAddSingleton in WalletContextServices does not override.
+        platformServices?.Invoke(services);
+        if (!services.Any(d => d.ServiceType == typeof(ISecureKeyProvider)))
+        {
+            if (OperatingSystem.IsWindows())
+                services.AddSingleton<ISecureKeyProvider, DpapiSecureKeyProvider>();
+            else if (OperatingSystem.IsLinux())
+                services.AddSingleton<ISecureKeyProvider, LinuxSecureKeyProvider>();
+            else
+                services.AddSingleton<ISecureKeyProvider, InMemorySecureKeyProvider>();
```
dangershony: What's the inmemory for? this is probably a bad idea no because the key will vanish once app is closed

src/sdk/Angor.Sdk/Common/FieldEncryption.cs
```diff
@@ -0,0 +1,55 @@
+using System.Security.Cryptography;
+using System.Text;
+
+namespace Angor.Sdk.Common;
+
+/// <summary>
+/// Lightweight AES-GCM encryption for individual string fields.
+/// Uses the raw key directly (no key derivation) since keys from ISecureKeyProvider are already cryptographically random.
+/// </summary>
+public static class FieldEncryption
```
dangershony, line 10: This is pretty standard code right?

src/sdk/Angor.Sdk/Common/WalletAccountBalanceService.cs
```diff
@@ -106,4 +127,40 @@ public async Task<Result> DeleteAccountBalanceInfoAsync(WalletId walletId)
...
+        var maybeKey = await secureKeyProvider.Get(walletId);
+        if (maybeKey.HasNoValue)
+        {
+            logger.LogWarning("No encryption key found for wallet {WalletId}, storing ExtPubKeys unencrypted", walletId);
+            return;
+        }
+
+        var key = maybeKey.Value;
+
+        if (!string.IsNullOrEmpty(accountInfo.ExtPubKey))
```
dangershony: this two lines are a bit odd they encrypt themselves? why do it this way?

src/sdk/Angor.Sdk.Tests/Common/InMemorySecureKeyProvider.cs
```diff
@@ -0,0 +1,37 @@
+using System.Security.Cryptography;
+using Angor.Sdk.Common;
+using Angor.Sdk.Wallet.Infrastructure.Interfaces;
+using CSharpFunctionalExtensions;
+
+namespace Angor.Sdk.Wallet.Infrastructure.Impl;
+
+public class InMemorySecureKeyProvider : ISecureKeyProvider
```
dangershony, line 8: Can you explain why we need this class?

src/sdk/Angor.Sdk/Wallet/Infrastructure/Impl/SecureKeyPasswordProvider.cs
```diff
@@ -0,0 +1,13 @@
+using Angor.Sdk.Common;
+using Angor.Sdk.Wallet.Infrastructure.Interfaces;
+using CSharpFunctionalExtensions;
+
+namespace Angor.Sdk.Wallet.Infrastructure.Impl;
+
+public class SecureKeyPasswordProvider(ISecureKeyProvider secureKeyProvider) : IPasswordProvider
```
dangershony, line 7: what is the purpose of this class?

src/shared/Angor.Shared/WalletOperations.cs
```diff
@@ -470,12 +470,12 @@ public AccountInfo BuildAccountInfoForWalletWords(WalletWords walletWords)
             _hdOperations.GetExtendedPublicKey(privateKey, extendedKey.ChainCode, accountHdPath);
 
         var rootExtPubKey = extendedKey.Neuter();
-        
-        var rootExtPubKeyHash = Convert.ToHexString(Hashes.SHA256(rootExtPubKey.ToBytes()));
+
+        var walletIdHash = Convert.ToHexString(Hashes.SHA256(accountExtPubKeyTostore.ToBytes()));
```
dangershony: this will make the wallet id per network right?

src/design/App/UI/Sections/FindProjects/InvestPageViewModel.cs
```diff
@@ -503,7 +661,26 @@ private static DateTimeOffset ComputeSpecificDayOfWeek(DateTimeOffset startDate,
...
+    /// <summary>
+    /// Auto-create a wallet if none exists. Used by the 1-click invest flow so the user
+    /// doesn't need to visit the Funds section before investing.
+    /// </summary>
+    private async Task<Result> EnsureWalletExistsAsync()
+    {
+        _logger.LogInformation("No wallet found — auto-creating for 1-click invest flow");
+        PaymentStatusText = "Creating wallet...";
 
+        var result = await _walletAppService.CreateWallet(_getNetwork() + " Wallet", _getNetwork());
```
dangershony: not to clear what this is about

src/sdk/Angor.Sdk/Common/WalletAccountBalanceService.cs
```diff
@@ -8,25 +9,31 @@ namespace Angor.Sdk.Common;
...
     public async Task<Result<AccountBalanceInfo>> GetAccountBalanceInfoAsync(WalletId walletId)
     {
         var result = await collection.FindByIdAsync(walletId.Value);
         if (result.IsFailure || result.Value is null)
             return Result.Failure<AccountBalanceInfo>("Account balance not found. Please refresh your wallet.");
-        
-        return Result.Success(result.Value.AccountBalanceInfo);
+
+        var accountBalanceInfo = result.Value.AccountBalanceInfo;
+        await DecryptExtPubKeys(walletId, accountBalanceInfo.AccountInfo);
```
dangershony, line 23: this means every time we interact with account balance we encrypt and decrypt?

dangershony:
Come to think about it I am not sure we want the wallet to be network aware, if we want a wallet to be able to show projects from different networks then we need to not make the network separation. the same wallet can have mainnet and liquid projects and the distinction is done in the ProjectInfo (we already have that).

I suggest you make a separate pr for the wallet per network and we can push the secure storage changes

src/sdk/Angor.Sdk/Common/WalletAccountBalanceService.cs
DavidGershony: Yes. but do we do it that often?

src/sdk/Angor.Sdk/Common/WalletAccountBalanceService.cs
DavidGershony: Yea.. it overrides the value going in to the db and back out. I wasn't sure if we should have a separate field here that is null in the db for that or keep it like that.

src/sdk/Angor.Sdk/Wallet/Infrastructure/Impl/SecureKeyPasswordProvider.cs
DavidGershony: We use the existing flow and existing interface, the injected ISecureKeyProvider is implemented differently for each OS

src/sdk/Angor.Sdk/Common/WalletAccountBalanceService.cs
dangershony: I see normally a separate field is less confusing I guess, but this sis more efficient so up to you.

src/sdk/Angor.Sdk/Common/FieldEncryption.cs
DavidGershony: Yes. we use the aesGcm class from system.security.cryptography to encrypt/decrypt

dangershony (approved):

PR #848, opened 2026-05-13
feat: resume pending lightning swap on app reopen

DavidGershony:
## Summary

- Detect and resume interrupted lightning swaps when the payment flow opens, preventing loss of funds locked on-chain by Boltz
- Add `GetResumableSwapsAsync` to `IBoltzSwapStorageService` to find swaps in non-terminal states (not claimed/failed/expired)
- Store `RequestedAmountSats` on `BoltzSwapDocument` so pending swaps can be matched to new payment flows by amount
- If the invoice was already paid (funds locked on-chain), the swap is always resumed regardless of amount mismatch to avoid fund loss

## Changes

- `PaymentFlowViewModel`: on Lightning tab init, check for pending swaps before starting a new one; restore UI state and resume monitoring
- `BoltzSwapStorageService` (SDK + Blazor): add `GetResumableSwapsAsync`, `TerminalStatuses` set, updated `SaveSwapAsync` signature
- `IBoltzSwapStorageService`: new `GetResumableSwapsAsync` method, updated `SaveSwapAsync` with `requestedAmountSats` param
- `BoltzSwapDocument`: add `RequestedAmountSats` field, updated `FromSwapModel` factory
- `CreateLightningSwap`: pass `request.Amount.Sats` to storage
- `InvestPageViewModel`, `DeployFlowViewModel`, `CompositionRoot`: wire through `IBoltzSwapStorageService` dependency
- Tests: updated Moq setups for new `SaveSwapAsync` signature


src/design/App/UI/Sections/FindProjects/InvestPageViewModel.cs
```diff
@@ -106,17 +106,17 @@ public bool IsSelected
 /// Supports three project types: Invest, Fund, Subscription.
 /// Subscription type shows plan cards (3mo/6mo) instead of BTC amount input.
 /// </summary>
-public partial class InvestPageViewModel : ReactiveObject, IDisposable
+public partial class InvestPageViewModel : ReactiveObject
 {
     private readonly IWalletAppService _walletAppService;
     private readonly IInvestmentAppService _investmentAppService;
     private readonly IBoltzSwapService _boltzSwapService;
+    private readonly IBoltzSwapStorageService _swapStorageService;
```
dangershony, line 114: _swapStorageService was added but it is never used (not a big deal though) in this vm

src/sdk/Angor.Sdk/Integration/Lightning/BoltzSwapStorageService.cs
```diff
@@ -13,11 +13,11 @@ public class BoltzSwapStorageService(
     IGenericDocumentCollection<BoltzSwapDocument> collection,
     ILogger<BoltzSwapStorageService> logger) : IBoltzSwapStorageService
 {
-    public async Task<Result> SaveSwapAsync(BoltzSubmarineSwap swap, string walletId, string? projectId = null)
+    public async Task<Result> SaveSwapAsync(BoltzSubmarineSwap swap, string walletId, string? projectId = null, long requestedAmountSats = 0)
```
dangershony, line 16: this defaults to zero `long requestedAmountSats = 0` is that correct wont we always have an amount?

src/webapp/Angor.Client/Storage/BoltzSwapStorageService.cs
```diff
@@ -204,6 +204,37 @@ public Task<Result> MarkSwapClaimedAsync(string swapId, string walletId, string
         }
     }
 
+    private static readonly HashSet<string> TerminalStatuses = new(StringComparer.OrdinalIgnoreCase)
```
dangershony, line 207: I see there are two BoltzSwapStorageService.cs (one for the web)

dangershony (approved):

dangershony:
running integration tests now

dangershony:
Two comments to consider
- when I move to the onchain tab and back to the lightning it create a new lightning swap (the one saved is not preserved)
- perhaps if the user lands on the page we should already let the user know there is an active swap and instead of Invest Now we make the button Continue Swap or something

PR #851, opened 2026-05-13
fix: non-breaking security hardening (C1, H2, H3, M2-M8)

DavidGershony:
## Summary

Implements all **non-breaking** and **backward-compatible** fixes from the security audit (`SECURITY_ANALYSIS.md`). These changes harden existing behavior without altering on-chain formats, derivation paths, or protocol semantics — safe to ship without protocol versioning.

## Changes by finding

| ID | Severity | Title | Fix |
|----|----------|-------|-----|
| **C1** | Critical | Sensitive data in logs | Stripped signatures/sighashes from `Information`-level logs (5 sites, 3 files) |
| **H2** | High | Missing key-ownership validation (founder) | Added `ScriptPubKey` mismatch check before signing in `WalletOperations` |
| **H3** | High | Missing key-ownership validation (investor/recovery) | Same check at all 3 signing call sites |
| **M2** | Medium | nSequence not signaling RBF | Changed `Sequence(spendingTrx.LockTime.Value)` to `Sequence(0xFFFFFFFE)` (2 sites) |
| **M3** | Medium | Fragile witness reconstruction | Replaced `Skip/Take` with explicit indexed loop + minimum push-count validation |
| **M4** | Medium | Missing recovery signature verification | Added `CheckRecoverySignatures` call at top of `AddSignaturesToRecoveryPathTransaction` |
| **M5** | Medium | `Debug.Assert` for signature verification | Replaced with `throw new InvalidOperationException` |
| **M6** | Medium | Unbounded seeder combinations | Added `BinomialCoefficient` guard capping threshold combinations at 10,000 |
| **M7** | Medium | Implicit endianness assumption | Replaced `BitConverter.GetBytes` with explicit `BinaryPrimitives.WriteInt16LittleEndian` |
| **M8** | Medium | Weak OP_RETURN parsing | Added pubkey length validation (33 bytes), null checks, explicit error messages |

## Files changed (7 files, +86 -22)

- `FounderTransactionActions.cs` — C1, M5
- `InvestorTransactionActions.cs` — C1, M3, M4
- `SeederTransactionActions.cs` — C1
- `ProjectScriptsBuilder.cs` — M7, M8
- `SeederScriptTreeBuilder.cs` — M6
- `SpendingTransactionBuilder.cs` — M2
- `WalletOperations.cs` — H2, H3

## Testing

- **146 shared tests** — all pass
- **312 SDK tests** — all pass
- No new tests added (hardening of existing paths; existing tests cover the happy paths)

## What's NOT in this PR (deferred)

- **C2/C3** (mnemonic memory protection + private key type-safety refactor) — separate PR, ~50 call sites
- **Breaking protocol changes** (H1, H4, H5, M1, L2) — require `ProjectInfo.Version` bump to v3 with dual-version support

### DIFF
--- a/src/shared/Angor.Shared/Protocol/InvestorTransactionActions.cs
+++ b/src/shared/Angor.Shared/Protocol/InvestorTransactionActions.cs
@@ -208,18 +208,23 @@ public class InvestorTransactionActions : IInvestorTransactionActions
             },
             (witScript, sig) =>
             {
+                if (witScript.PushCount < 3)
+                    throw new InvalidOperationException($"Unexpected witness structure: expected at least 3 pushes (sig, script, controlblock), got {witScript.PushCount}");
+
                 var controBlock = new NBitcoin.Script(witScript[witScript.PushCount - 1]);
                 var scriptToExecute = new NBitcoin.Script(witScript[witScript.PushCount - 2]);
 
                 List<Op> ops = new List<Op>();
 
-                // the last 3 items on the stack are the fakesig, script and controlblock anything before that is the secrets
+                // Witness layout: [fakeSig] [secret_0 .. secret_n] [scriptToExecute] [controlBlock]
+                // Replace fakeSig with real sig, preserve secrets, reattach script + controlblock
+                var secretCount = witScript.PushCount - 3; // exclude fakeSig, script, controlblock
 
                 ops.Add(Op.GetPushOp(sig.ToBytes()));
 
-                foreach (var oppush in witScript.Pushes.Skip(1).Take(witScript.Pushes.Count() - 3))
+                for (int i = 0; i < secretCount; i++)
                 {
-                    ops.Add(Op.GetPushOp(oppush));
+                    ops.Add(Op.GetPushOp(witScript[i + 1])); // secrets start at index 1
                 }
 
                 ops.Add(Op.GetPushOp(scriptToExecute.ToBytes()));
@@ -282,6 +287,10 @@ public class InvestorTransactionActions : IInvestorTransactionActions
 
     private Transaction AddSignaturesToRecoveryPathTransaction(ProjectInfo projectInfo, Transaction investmentTransaction, Transaction recoveryTransaction, SignatureInfo founderSignatures, string investorPrivateKey)
     {
+        // Verify founder signatures before incorporating them
+        if (!CheckRecoverySignatures(projectInfo, investmentTransaction, recoveryTransaction, founderSignatures))
+            throw new InvalidOperationException("Founder recovery signature verification failed. Refusing to co-sign with invalid founder signatures.");
+
         var (investorKey, secretHash) = _projectScriptsBuilder.GetInvestmentDataFromOpReturnScript(investmentTransaction.Outputs.First(_ => _.ScriptPubKey.IsUnspendable).ScriptPubKey);
 
         var nbitcoinNetwork = NetworkMapper.Map(_networkConfiguration.GetNetwork());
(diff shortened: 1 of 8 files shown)


src/shared/Angor.Shared/Protocol/Scripts/ProjectScriptsBuilder.cs
```diff
@@ -88,27 +104,38 @@ public Script BuildSeederInfoScript(ProjectInfo projectInfo, FundingParameters p
...
                 return (pubKey.ToHex(), secretHash);
             }
-            else if (ops[2].PushData?.Length == 4)
+            else if (ops[2].PushData.Length == 4)
             {
                 // Dynamic stage info (no secret hash)
                 return (new PubKey(ops[1].PushData).ToHex(), null);
             }
+            else
+            {
```
dangershony, line 126: yeah this totally makes sense, I think we have another place that we parse this that we should probably unify the code

src/shared/Angor.Shared/Protocol/Scripts/SeederScriptTreeBuilder.cs
```diff
@@ -39,12 +39,34 @@ public IEnumerable<Script> BuildSeederScriptTree(string investorKey, int seederT
     
     public static List<List<string>> CreateThresholds(int threshold, List<string> secretHashes)
     {
+        // Guard against combinatorial explosion: C(n,k) can grow very large.
+        // Cap at a reasonable limit to prevent OOM/DoS.
+        const int maxCombinations = 10_000;
```
dangershony, line 44: this code is not used right now right?

src/shared/Angor.Shared/Protocol/TransactionBuilders/SpendingTransactionBuilder.cs
```diff
@@ -63,7 +63,7 @@ public TransactionInfo BuildRecoverInvestorRemainingFundsInProject(string invest
 
              return new TxIn(new OutPoint(_.Transaction, _.N))
              {
-                 Sequence = new Sequence(spendingTrx.LockTime.Value),
+                 Sequence = new Sequence(0xFFFFFFFE),
```
dangershony, line 66: this changes breaks the Angor protocol if I am not mistaken

The difference is in what Sequence is set to:
1. new Sequence(0xFFFFFFFE) — This is a fixed constant meaning "opt-in to nLockTime but don't use RBF (replace-by-fee)." It simply signals that the transaction's LockTime field should be enforced by consensus, without encoding any relative timelock in the input itself.
2. new Sequence(spendingTrx.LockTime.Value) — This encodes the transaction's LockTime value directly into the input's Sequence field. This is used for relative timelocks (BIP 68 / OP_CHECKSEQUENCEVERIFY), where the Sequence field itself specifies how many blocks/time must pass since the UTXO was confirmed before it can be spent.
In short: the first says "respect the transaction-level locktime"; the second says "enforce a relative timelock equal to LockTime on this specific input." They serve different script spending paths — one uses OP_CHECKLOCKTIMEVERIFY (absolute), the other uses OP_CHECKSEQUENCEVERIFY (relative).

dangershony: ok after more consideration I think this is ok 0xFFFFFFFE is the standard value to enable CLTV

src/shared/Angor.Shared/Protocol/InvestorTransactionActions.cs
```diff
@@ -282,6 +287,10 @@ public bool IsInvestmentAbovePenaltyThreshold(ProjectInfo projectInfo, long inve
 
     private Transaction AddSignaturesToRecoveryPathTransaction(ProjectInfo projectInfo, Transaction investmentTransaction, Transaction recoveryTransaction, SignatureInfo founderSignatures, string investorPrivateKey)
     {
+        // Verify founder signatures before incorporating them
+        if (!CheckRecoverySignatures(projectInfo, investmentTransaction, recoveryTransaction, founderSignatures))
```
dangershony, line 291: I am pretty sure we do it at a higher level but no harm in doing  more checks

dangershony (approved):

PR #854, opened 2026-05-13
security: AngorKey type-safety (C3) and mnemonic memory protection (C2)

DavidGershony:
## Summary

Implements two critical security findings from the security audit:

- **C3 (Private key type-safety)**: Introduces `AngorKey : NBitcoin.Key` to replace raw hex string private key parameters across all protocol interfaces, eliminating accidental logging/serialization of key material
- **C2 (Mnemonic memory protection)**: Refactors `WalletWords` to use `char[]` backing with `IDisposable` zeroing via `CryptographicOperations.ZeroMemory`, caches the derived `ExtKey` to avoid repeated PBKDF2 derivation, and updates `SensitiveWalletDataProvider` to deterministically zero cached mnemonics on removal

### C3 - AngorKey type-safety
- New `AngorKey : NBitcoin.Key` class with `FromHex(string)` and `From(Blockcore.NBitcoin.Key)` bridge factories
- Updated 7 protocol interfaces + implementations to accept `AngorKey` instead of `string`
- Updated 8 SDK operation handlers and 7 webapp Razor pages
- Updated all test files (Shared + SDK)

### C2 - Mnemonic memory protection
- `WalletWords` rewritten with `char[]` backing, `IDisposable` zeroing, and lazy `ExtKey` caching
- `DerivationOperations` routes all 12+ derivation sites through `walletWords.GetOrDeriveExtKey()`
- `WalletOperations.DerivePublicKey/DerivePrivateKey` now use cached `ExtKey` overloads
- `IHdOperations` extended with `DerivePublicKey(ExtKey, string)` and `DerivePrivateKey(ExtKey, string)` overloads
- `SensitiveWalletDataProvider` caches `WalletWords` (zeroable) instead of raw `(string, Maybe<string>)` tuples, with `Dispose()` zeroing on removal

## Test Results

- Shared tests: **146/146 passed**
- SDK tests: **312/312 passed** (7 skipped - pre-existing)
- Builds clean (only pre-existing iOS workload error on Windows)

### DIFF
--- a/src/sdk/Angor.Sdk/Wallet/Infrastructure/Impl/SensitiveWalletDataProvider.cs
+++ b/src/sdk/Angor.Sdk/Wallet/Infrastructure/Impl/SensitiveWalletDataProvider.cs
@@ -1,36 +1,75 @@
 using Angor.Sdk.Common;
 using Angor.Sdk.Wallet.Domain;
 using Angor.Sdk.Wallet.Infrastructure.Interfaces;
+using Angor.Shared.Models;
 using CSharpFunctionalExtensions;
 using CSharpFunctionalExtensions.ValueTasks;
 
 namespace Angor.Sdk.Wallet.Infrastructure.Impl;
 
+/// <summary>
+/// Caches wallet sensitive data (mnemonic words, passphrase) using WalletWords
+/// which backs storage with char[] arrays that can be zeroed on disposal.
+/// </summary>
 public class SensitiveWalletDataProvider(IWalletStore walletStore, IWalletSecurityContext walletSecurityContext,
-    IWalletEncryption walletEncryption) : ISensitiveWalletDataProvider
+    IWalletEncryption walletEncryption) : ISensitiveWalletDataProvider, IDisposable
 {
-    private readonly Dictionary<WalletId, (string, Maybe<string>)> cachedSensitiveData = new();
+    private readonly Dictionary<WalletId, WalletWords> cachedWalletWords = new();
 
     public async Task<Result<(string seed, Maybe<string> passphrase)>> RequestSensitiveData(WalletId walletId)
     {
-        var findResult = cachedSensitiveData.TryFind(walletId);
-        if (findResult.HasValue)
+        if (cachedWalletWords.TryGetValue(walletId, out var cached))
         {
-            return findResult.Value;
+            return ToTuple(cached);
         }
 
-        var result = await RequestSensitiveDataCore(walletId).Tap(data => cachedSensitiveData[walletId] = data);
+        var result = await RequestSensitiveDataCore(walletId);
+        if (result.IsSuccess)
+        {
+            var walletWords = new WalletWords
+            {
+                Words = result.Value.seed,
+                Passphrase = result.Value.passphrase.GetValueOrDefault("")
+            };
+            cachedWalletWords[walletId] = walletWords;
+        }
+
         return result;
     }
 
     public void SetSensitiveData(WalletId id, (string seed, Maybe<string> passphrase) data)
     {
-        cachedSensitiveData[id] = data;
+        // Dispose any previously cached data for this wallet
+        if (cachedWalletWords.TryGetValue(id, out var existing))
+            existing.Dispose();
+
+        cachedWalletWords[id] = new WalletWords
+        {
+            Words = data.seed,
+            Passphrase = data.passphrase.GetValueOrDefault("")
+        };
     }
 
     public void RemoveSensitiveData(WalletId id)
     {
-        cachedSensitiveData.Remove(id);
+        if (cachedWalletWords.Remove(id, out var walletWords))
+            walletWords.Dispose();
+    }
+
+    public void Dispose()
+    {
+        foreach (var walletWords in cachedWalletWords.Values)
+            walletWords.Dispose();
+
+        cachedWalletWords.Clear();
+    }
+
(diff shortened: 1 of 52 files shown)


src/sdk/Angor.Sdk/Wallet/Infrastructure/Impl/SensitiveWalletDataProvider.cs
```diff
@@ -1,36 +1,75 @@
...
 
+/// <summary>
+/// Caches wallet sensitive data (mnemonic words, passphrase) using WalletWords
+/// which backs storage with char[] arrays that can be zeroed on disposal.
+/// </summary>
 public class SensitiveWalletDataProvider(IWalletStore walletStore, IWalletSecurityContext walletSecurityContext,
-    IWalletEncryption walletEncryption) : ISensitiveWalletDataProvider
+    IWalletEncryption walletEncryption) : ISensitiveWalletDataProvider, IDisposable
 {
-    private readonly Dictionary<WalletId, (string, Maybe<string>)> cachedSensitiveData = new();
```
dangershony, line 12: I am so happy to get rid of this weird `Maybe<>` type never understood it

src/sdk/Angor.Sdk/Wallet/Infrastructure/Impl/SensitiveWalletDataProvider.cs
```diff
@@ -1,36 +1,75 @@
...
+    private readonly Dictionary<WalletId, WalletWords> cachedWalletWords = new();
```
dangershony, line 17: so we cache the words but the type is disposable

src/sdk/Angor.Sdk/Wallet/Infrastructure/Impl/SensitiveWalletDataProvider.cs
```diff
@@ -1,36 +1,75 @@
...
+    {
+        foreach (var walletWords in cachedWalletWords.Values)
+            walletWords.Dispose();
+
+        cachedWalletWords.Clear();
+    }
+
+    private static (string seed, Maybe<string> passphrase) ToTuple(WalletWords walletWords)
+    {
+        Maybe<string> passphrase = string.IsNullOrEmpty(walletWords.Passphrase)
```
dangershony, line 69: do we need Maybe type?

src/shared/Angor.Shared/Models/WalletWords.cs
```diff
@@ -1,11 +1,66 @@
...
+
+    /// <summary>The BIP-39 mnemonic words (space-separated).</summary>
+    public string Words
+    {
+        get
+        {
+            ObjectDisposedException.ThrowIf(disposed, this);
+            return new string(wordsChars);
+        }
+        set => wordsChars = value?.ToCharArray() ?? Array.Empty<char>();
```
dangershony: Shouldn't this throw? it means we are setting null.
Will we ever set the words to be null ? (hmm maybe)

src/shared/Angor.Shared/Models/WalletWords.cs
```diff
@@ -19,4 +74,17 @@ public static WalletWords ConvertFromString(string data)
...
     }
-}
\ No newline at end of file
+
+    public void Dispose()
+    {
+        if (disposed) return;
+        disposed = true;
+
+        CryptographicOperations.ZeroMemory(MemoryMarshal.AsBytes(wordsChars.AsSpan()));
```
dangershony, line 88: I didn't know about ZeroMemory, fantastic addition to security

src/shared/Angor.Shared/DerivationOperations.cs
```diff
@@ -225,7 +208,7 @@ public async Task<Key> DeriveProjectNostrPrivateKeyAsync(WalletWords walletWords
 
         var extKey = await Task.Run(() => extendedKey.Derive(new KeyPath(path)));
         
-        return extKey.PrivateKey;
+        return AngorKey.From(extKey.PrivateKey);
     }
 
     public uint DeriveUniqueProjectIdentifier(string founderKey)
```
dangershony, line 214: weren't we suppose to make a change in this method as well to increase the entropy of the project id derivation?

dangershony (dismissed):

dangershony:
I got an error here of invalid encryption key which should not happen so I am investigating dont commit yet

src/sdk/Angor.Sdk/Wallet/Infrastructure/Impl/SensitiveWalletDataProvider.cs
DavidGershony: Yes, we were already caching but you can't zero the value of the string

src/sdk/Angor.Sdk/Wallet/Infrastructure/Impl/SensitiveWalletDataProvider.cs
DavidGershony: We'll replace it with nullable when removing the library

src/shared/Angor.Shared/Models/WalletWords.cs
DavidGershony: Claude agreed with you :) - fixed

src/shared/Angor.Shared/DerivationOperations.cs
DavidGershony: That should happen in another PR.

dangershony (approved):

PR #855, opened 2026-05-13
fix: prevent wallet loss on Android swipe-kill

dangershony:
## Summary
- Add `OnStop` lifecycle handler in `MainActivity` to checkpoint LiteDB (flush WAL to main data file) and immediately persist prototype settings (selected wallet ID, theme) before Android can kill the process
- Add WAL recovery in `LiteDbDocumentDatabase` — if the database fails to open due to a corrupted journal from an unclean shutdown, the `-log` WAL file is deleted and the database is retried, preserving the main data file
- Add `FlushAsync()` to `PrototypeSettings` for immediate persistence, bypassing the 250ms throttle used during normal operation

## Root Cause
When the user swipes the app away on Android, the process is killed without any lifecycle hooks firing to flush pending writes. This could cause:
1. **LiteDB WAL corruption** — if a write was in-flight, the WAL journal could be left in an inconsistent state, causing `GetMetadatas()` to fail on next launch and the wallet list to appear empty
2. **Lost `SelectedWalletId`** — the settings save was throttled by 250ms and ran on a background task, so it could be dropped if the process was killed too quickly

## Files Changed
- `src/design/App.Android/MainActivity.cs` — `OnStop` override
- `src/design/App/UI/Shell/ShellViewModel.cs` — `PrototypeSettings.FlushAsync()`
- `src/sdk/Angor.Data.Documents.LiteDb/LiteDbDocumentDatabase.cs` — `OpenDatabaseWithRecovery()`

### DIFF
--- a/src/design/App.Android/MainActivity.cs
+++ b/src/design/App.Android/MainActivity.cs
@@ -1,10 +1,15 @@
+using System;
 using Android.App;
 using Android.Content;
 using Android.Content.PM;
 using Android.Views;
+using Angor.Data.Documents.Interfaces;
 using App.UI.Shared.Helpers;
+using App.UI.Shell;
 using Avalonia.Android;
 using Avalonia.Threading;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 
 namespace App.Android;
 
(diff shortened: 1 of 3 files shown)


src/design/App.Android/MainActivity.cs
```diff
@@ -39,6 +44,44 @@ protected override void OnPause()
...
+
+        // Flush all pending state to disk before the process may be killed.
+        // Android can terminate the process at any point after OnStop returns.
+        try
+        {
+            var services = global::App.App.Services;
+            if (services == null) return;
+
+            // Checkpoint LiteDB to flush the WAL into the main data file
+            var db = services.GetService<IAngorDocumentDatabase>();
```
DavidGershony, line 59: I think we should inject this interface and not pull the full dervice provider from global for it

src/sdk/Angor.Data.Documents.LiteDb/LiteDbDocumentDatabase.cs
```diff
@@ -177,4 +168,43 @@ public void Dispose()
...
+            return db;
+        }
+        catch (Exception ex)
+        {
+            _logger.LogWarning(ex, "LiteDB failed to open, attempting WAL recovery: {Path}", _databasePath);
+
+            try
+            {
+                // LiteDB 5.x WAL file uses the "-log" suffix
+                var walPath = _databasePath + "-log";
```
DavidGershony, line 192: So we delete a corrupt db and start over?

dangershony: only the log, not the database

dangershony:
I pushed a commit that injects `ILogger<PrototypeSettings>` via the constructor instead of resolving it from `App.Services` in both `ScheduleSaveSettings` and `FlushAsync`.

Regarding `MainActivity.OnStop()` — it still uses `App.Services` (the global service provider) and that's unavoidable: `MainActivity` is instantiated by the Android runtime, not by our DI container, so we can't constructor-inject dependencies into it. The service locator pattern is the standard approach for Android lifecycle callbacks.

src/design/App.Android/MainActivity.cs
dangershony: You cannot inject service in the android version so this is is the standard way, but it fixed the logger in the flush method

PR #869, opened 2026-05-18
Advanced Editor for Investment Stages + UTC Date Fix

mistic0xb:
**Issues fixed:**

1. "Advanced Editor" button in Step 5 did nothing — no UI existed behind the toggle
2. Project dates (start, end, stage release) stored 1 day behind for users in non-UTC timezones

 
**What changed:**
1. Advanced Editor (Create Project -> Step 5)
2. Built the Advanced Editor UI: editable stage rows with NumericUpDown for %, CalendarDatePicker for release date, add/remove stage buttons
3. Live "Total: X% — needs Y% more" badge updates as user edits percentages
4. Inline error shown when a stage has 0% or a date before the funding end date
5. New stages default to the remaining % needed to reach 100 instead of 0
6. Removing all stages via the editor re-shows the generate form so the user isn't stuck
7. UTC timezone fix
8. Root cause: DateTime.TryParse and DateOnly.ToDateTime(TimeOnly.MinValue) both produce DateTimeKind.Unspecified; the JSON serializer calls .ToUniversalTime() which treats Unspecified as local time and subtracts the UTC offset — shifting all dates back by the user's UTC offset
9. Fixed by explicitly setting DateTimeKind.Utc on StartDate, EndDate, and all stage ReleaseDate values before passing to the SDK

### DIFF
--- a/src/design/App/UI/Sections/MyProjects/CreateProjectViewModel.cs
+++ b/src/design/App/UI/Sections/MyProjects/CreateProjectViewModel.cs
@@ -330,7 +330,7 @@ public partial class CreateProjectViewModel : ReactiveObject
                 this.RaisePropertyChanged(nameof(HasFormError));
                 this.RaisePropertyChanged(nameof(HasEndDateError));
                 // Revalidate all stage release dates against the new end date
-                foreach (var s in Stages) UpdateStageDateError(s);
+                UpdateAllStageDateErrors();
             });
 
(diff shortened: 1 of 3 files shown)


dangershony:
@mistic0xb I assume you tested this?

dangershony:
image

This seems to be an error

mistic0xb:
looking at it

dangershony (approved):

PR #876, opened 2026-05-21
fix: add WebSocket reconnect logic for Boltz swap monitoring

dangershony:
## Problem

When the Android app goes to background (OnPause/OnStop), the OS kills all network sockets. The Boltz WebSocket monitoring a Lightning swap dies with \ConnectionClosedPrematurely_Generic\ and never reconnects, leaving the user stuck on 'Waiting for Lightning payment...' with a WebSocket error banner.

**Evidence from logs (angor-20260522.log):**
- \

### DIFF
--- a/src/shared/Angor.Shared/Integration/Lightning/BoltzWebSocketClient.cs
+++ b/src/shared/Angor.Shared/Integration/Lightning/BoltzWebSocketClient.cs
@@ -38,6 +38,9 @@ public class BoltzWebSocketClient : IBoltzWebSocketClient, IAsyncDisposable
         };
     }
 
+    private const int MaxReconnectAttempts = 5;
+    private static readonly TimeSpan InitialReconnectDelay = TimeSpan.FromSeconds(1);
+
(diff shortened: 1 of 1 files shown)


src/shared/Angor.Shared/Integration/Lightning/BoltzWebSocketClient.cs
```diff
@@ -138,7 +176,7 @@ private async Task<Result<BoltzSwapStatus>> ReceiveUpdatesUntilComplete(
             }
         }
 
-        return Result.Failure<BoltzSwapStatus>("WebSocket connection ended unexpectedly");
+        throw new WebSocketException("WebSocket connection ended unexpectedly");
```
DavidGershony: Why are we throwing and not returning a result?

DavidGershony (dismissed):

src/shared/Angor.Shared/Integration/Lightning/BoltzWebSocketClient.cs
dangershony: fixed

PR #895, opened 2026-06-10
ci: add portable (no-install) Windows exe to release artifacts

dangershony:
﻿Adds standalone portable exe builds for Windows x64 and arm64 alongside the existing Setup installers.

## What this does

Each release will now produce two additional Windows artifacts:

- `Angor-{version}-win-x64-Portable.exe` - single self-contained exe, no install needed
- `Angor-{version}-win-arm64-Portable.exe` - single self-contained exe, no install needed

These use `dotnet publish` with `--self-contained` and the project's existing `PublishSingleFile`/`PublishTrimmed` settings, so users just download and double-click - no setup wizard required.

## Context

Currently only Setup installers are produced. Having a no-install portable option is useful for users who prefer not to run a setup wizard.

### DIFF
--- a/.github/workflows/release-avalonia.yml
+++ b/.github/workflows/release-avalonia.yml
@@ -99,6 +99,16 @@ jobs:
             --appId ${{ env.APP_ID }} `
             --version ${{ needs.prepare.outputs.version }} `
             --vendor "${{ env.VENDOR }}"
+
+      - name: Create Windows x64 Portable
+        run: |
+          dotnet publish ${{ env.DESKTOP_PROJECT }} -c Release -r win-x64 --self-contained -p:Version=${{ needs.prepare.outputs.version }} -o publish-win-x64
+          Copy-Item publish-win-x64/Angor2.exe artifacts/Angor-${{ needs.prepare.outputs.version }}-win-x64-Portable.exe
+
+      - name: Create Windows arm64 Portable
+        run: |
+          dotnet publish ${{ env.DESKTOP_PROJECT }} -c Release -r win-arm64 --self-contained -p:Version=${{ needs.prepare.outputs.version }} -o publish-win-arm64
+          Copy-Item publish-win-arm64/Angor2.exe artifacts/Angor-${{ needs.prepare.outputs.version }}-win-arm64-Portable.exe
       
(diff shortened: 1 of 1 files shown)


dangershony:
## Signing with Azure Trusted Signing (next step)

To eliminate the SmartScreen "Unknown publisher" warning, we can add Azure Trusted Signing to this workflow. Here's the plan:

### Prerequisites

1. **Azure subscription** with a Trusted Signing account created (costs ~$10/month)
2. **Identity verification** completed in the Azure Trusted Signing portal
3. **Certificate profile** created (public trust, code signing)
4. Add these **GitHub secrets**:
   - `AZURE_CLIENT_ID`
   - `AZURE_TENANT_ID`
   - `AZURE_CLIENT_SECRET`
   - `AZURE_CODE_SIGNING_ACCOUNT`
   - `AZURE_CERTIFICATE_PROFILE`
   - `AZURE_ENDPOINT` (e.g. `https://eus.codesigning.azure.net`)

### Workflow change

Add a signing step after all Windows exes are built, before the upload step:

```yaml
- name: Sign Windows executables
  uses: azure/trusted-signing-action@v0.5.0
  with:
    azure-tenant-id: ${{ secrets.AZURE_TENANT_ID }}
    azure-client-id: ${{ secrets.AZURE_CLIENT_ID }}
    azure-client-secret: ${{ secrets.AZURE_CLIENT_SECRET }}
    endpoint: ${{ secrets.AZURE_ENDPOINT }}
    trusted-signing-account-name: ${{ secrets.AZURE_CODE_SIGNING_ACCOUNT }}
    certificate-profile-name: ${{ secrets.AZURE_CERTIFICATE_PROFILE }}
    files-folder: artifacts
    files-folder-filter: exe
    file-digest: SHA256
    timestamp-rfc3161: http://timestamp.acs.microsoft.com
    timestamp-digest: SHA256
```

This will sign all four exes (both Setup and Portable, both x64 and arm64) in a single step. The signed binaries get immediate SmartScreen trust -- no reputation buildup needed.

### Docs

- Azure Trusted Signing: https://learn.microsoft.com/en-us/azure/trusted-signing/overview
- GitHub Action: https://github.com/azure/trusted-signing-action

PR #912, opened 2026-07-03
H1: Add two-level BIP-32 derivation with 62-bit entropy for V3+ projects

DavidGershony:
## Summary

Add two-level BIP-32 hardened derivation paths for V3+ projects, increasing path entropy from 31 bits (~46k collision threshold) to 62 bits (~2 billion collision threshold).

## Problem

The current single-level derivation uses 31 bits of entropy from SHA-256d(founderKey), giving a Birthday-bound collision probability of ~46k projects before a 50% chance of path collision.

## Solution

- Add `DeriveProjectIndicesV2(founderKey)` that splits 62 bits into two 31-bit indices (each valid for BIP-32 hardened derivation, max 2^31-1)
- Add `BuildProjectSubPath(founderKey, projectVersion)` that returns:
  - V1/V2: `{upi}'` (single level, backward compatible)
  - V3+: `{hi}'/{lo}'` (two levels, 62-bit entropy)
- All derivation methods now have version-aware overloads defaulting to V1
- `DeriveAngorKey` also supports V3+ with two-level `Derive(hi).Derive(lo)`

## Version Gating

Uses `projectVersion` parameter (defaults to 1). Callers pass the project's version when creating V3+ projects. Existing V1/V2 projects continue using the single-level path.

## Testing

All 154 shared tests pass.


dangershony (approved):
When do we plan to hook the project version?
I also defaulted the version to 2 which is what we have on project info

PR #914, opened 2026-07-03
H4: Use BIP-341 NUMS point as unspendable Taproot internal key for V3+

DavidGershony:
## Summary

Use the well-known BIP-341 NUMS (Nothing Up My Sleeve) point as the Taproot internal key for V3+ projects, making the key path provably unspendable.

## Problem

The current implementation derives the internal key from project data. While it was intended to be unspendable, there's no cryptographic guarantee — a flaw in the derivation logic or a future refactor could accidentally make it spendable, allowing funds to bypass the script path.

## Solution

For V3+ projects, use the standard NUMS point from BIP-341:
`
0x50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0
`

This point is the SHA-256 hash of the Generator point G interpreted as a public key. No one can know its discrete logarithm, so the key path is cryptographically guaranteed to be unspendable.

## Changes

- Add `CreateUnspendableInternalKeyV2()` returning the NUMS `TaprootInternalPubKey`
- `BuildTaprootSpendInfo` now accepts `projectVersion` parameter
- V1/V2 continues using the legacy derived key (backward compatible)

## Testing

All 154 shared tests pass.

### DIFF
--- a/src/shared/Angor.Shared/Protocol/Scripts/TaprootScriptBuilder.cs
+++ b/src/shared/Angor.Shared/Protocol/Scripts/TaprootScriptBuilder.cs
@@ -87,9 +87,11 @@ public class TaprootScriptBuilder : ITaprootScriptBuilder
         return (new Script(controlBlock.ToBytes()), execute, secretHashes.ToArray());
     }
     
-    private static TaprootSpendInfo BuildTaprootSpendInfo(ProjectScripts scripts)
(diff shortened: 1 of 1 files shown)


dangershony (approved):

dangershony (approved):
We need to make sure the version is used in later upgrade right?

PR #915, opened 2026-07-03
H5: Use pubkey hash for storage paths (V3+) and zero sensitive bytes

DavidGershony:
## Summary

Avoid exposing private key bytes in derivation paths and ensure all sensitive byte arrays are zeroed after use.

### DeriveLeadInvestorSecretHash
- **V3+**: Hash only the derived **public key** — deterministic and doesn't require private key bytes at all
- **V1/V2**: Keep legacy behavior (hash full ExtKey serialization) but now zero the byte array in a `finally` block

### DeriveNostrStoragePassword
- Zero the private key byte array in a `finally` block after hashing
- Remove redundant `.Replace("-", "").ToLower()` (`Encoders.Hex.EncodeData` already returns lowercase hex without dashes)

## Rationale

Private key bytes left unzeroed in managed memory can be read by memory-scraping malware or appear in crash dumps. By zeroing them immediately after use and (for V3+) avoiding private key bytes entirely, we reduce the attack surface.

## Testing

All 154 shared tests pass. V1/V2 derivation produces the same hash output as before (backward compatible).


src/shared/Angor.Shared/DerivationOperations.cs
```diff
@@ -89,11 +94,26 @@ public string DeriveLeadInvestorSecretHash(WalletWords walletWords, string found
 
         ExtPubKey extPubKey = _hdOperations.GetExtendedPublicKey(extendedKey.PrivateKey, extendedKey.ChainCode, path);
 
-        var derivedSecret = extendedKey.Derive(new KeyPath(path));
-
-        var hash = Hashes.Hash256(derivedSecret.ToBytes()).ToString();
+        if (projectVersion >= 3)
+        {
+            // V3+: Hash only the public key (deterministic, doesn't expose private key bytes)
+            var pubKeyBytes = extPubKey.PubKey.ToBytes();
```
dangershony, line 97: this is deriving the secret hash, if we derive using the pubkey and the pubkey is leaked then anyone can derive the hash no?

dangershony: ah ok so basically this may not be an issue because we never reveal the pubkey anyway

src/shared/Angor.Shared/DerivationOperations.cs
```diff
@@ -77,6 +77,11 @@ public FounderKeys GetProjectKey(FounderKeyCollection founderKeyCollection, int
     }
 
     public string DeriveLeadInvestorSecretHash(WalletWords walletWords, string founderKey)
+    {
+        return DeriveLeadInvestorSecretHash(walletWords, founderKey, projectVersion: 1);
```
dangershony:
```suggestion
        return DeriveLeadInvestorSecretHash(walletWords, founderKey, projectVersion: 2);
```

PR #917, opened 2026-07-03
M8: Prepend version byte to V3+ OP_RETURN scripts

DavidGershony:
## Summary

Prepend a 1-byte version marker to OP_RETURN scripts for V3+ projects, enabling future protocol evolution while maintaining backward-compatible parsing.

## Problem

The current OP_RETURN format has no version indicator. If the data layout needs to change in the future, there's no way to distinguish between old and new formats during parsing — leading to silent misinterpretation of data.

### Builder (V3+)
Prepend a single byte with value equal to the project version (e.g., `0x03`) before the existing data pushes.

### Parser
Detect the version byte by checking:
1. First data push is exactly 1 byte long
2. Its value is >= 3

If detected, adjust the data start index to skip the version byte. Otherwise, parse using the legacy V1/V2 layout.

## Backward Compatibility

- V1/V2 projects have no version byte — parser recognizes them by the absence of the marker
- V3+ projects have the version byte — parser adjusts offsets accordingly
- No existing on-chain data is affected

## Testing

All 154 shared tests pass.

### DIFF
--- a/src/shared/Angor.Shared/Protocol/Scripts/ProjectScriptsBuilder.cs
+++ b/src/shared/Angor.Shared/Protocol/Scripts/ProjectScriptsBuilder.cs
@@ -32,8 +32,15 @@ public class ProjectScriptsBuilder : IProjectScriptsBuilder
         var ops = new List<Op>
         {
             OpcodeType.OP_RETURN,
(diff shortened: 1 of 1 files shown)


dangershony:
I dont think we need to add version to the opreturn because an opreturn is linked to a project info and there we have the verison already

PR #927, opened 2026-07-13
Funders live monitoring + badges, markdown descriptions, image perf, cache freshness + UI fixes

Zazawowow:
## Summary

Fixes three UI bugs in the Avalonia rewrite (`src/design`):

### 1. Funders amounts showing 0.0000 BTC
Direct (below-threshold) investments only carry a transaction ID, and the amount was looked up from the indexer's investment list, which often doesn't include them yet. `GetProjectInvestments` now falls back to fetching the raw transaction by ID and summing its taproot outputs when the indexed amount is 0.

### 2. Blank NIP-05 / private keys in Manage Project → Private Keys modal
- `Nip05` was never populated — now loaded from the project's Nostr profile metadata (`FetchProjectProfileData`). The nsec/hex fields use the `GetFounderNsec` operation from main.
- Fixed a mapping bug in `ProjectMetadata.Parse`: `Nip05` and `Lud16` were both incorrectly mapped from `Nip57`.

### 3. Off-center expand/collapse chevron in Funders
The chevron geometry (14×7 bounds in a 24×24 icon space) was stretched to fill its 16×16 viewbox. The four chevron `Path`s now use `Stretch="None"` so the icon keeps its centered position.

### 4. Live funders monitoring with badges + notifications (`0928be82`)
- New `FundersMonitor` singleton polls founder projects in the background (2 min interval) and keeps a warm snapshot — the Funders tab now renders instantly instead of scanning relays on every open
- Orange pending-approval count badge on the desktop Funders nav item and mobile Founder tab (previously a dead placeholder)
- Toasts on any tab when new funding requests arrive and when an approved funder completes their investment (approved → funded transition detection)
- Approved cards distinguish **Awaiting funding** vs **Funded**
- Fixed desktop tab entry not triggering refresh (`OnBecameActive` was only wired through the mobile `SectionPanel`)
- Full-page loader only before the first poll; later refreshes show data with the spinning refresh icon

### 5. Markdown rendering for project descriptions (`fcd911f4`)
- Dependency-free `MarkdownTextBlock`: headings, emphasis, lists (with lazy continuation), links, inline/fenced code, blockquotes, rules — themed via app tokens
- Media support: markdown images, inline HTML ``/`<video>`/`<a>`/`<br>`; video renders as an external-open card
- `MarkdownToPlainConverter` strips syntax in line-clamped summaries (cards, hero blurb, portfolio detail)

### 6. Scroll lag from oversized project logos (`932f01ce`)
`DecodeToWidth` infrastructure existed but was dead code — every banner/avatar decoded at full resolution, so a huge logo became a giant permanent texture resampled every scroll frame. All image loads now decode at bounded width (cards 640/128px, detail 1280/256px); cache keys include decode width.

### 7. Stale banners/avatars after profile updates (`23fa16bf`)
The LiteDB project cache was write-once — kind-0 profile metadata was never refetched. Now uses stale-while-revalidate with a 1h TTL and background relay revalidation; `UpdateProjectProfile` invalidates the founder's own cache immediately, and the project detail view applies fresh banner/picture URLs on open.

### Tests
- Updated `GetProjectInvestmentsHandler` test constructions for the new `IIndexerService` dependency
- Fixed stale `GetInvestmentsHandler` test constructions that were missing `IIndexerService` (pre-existing build break in the test project)
- All 323 SDK tests pass; App.Desktop builds clean


src/sdk/Angor.Sdk/Funding/Founder/Operations/UpdateProjectProfile.cs
```diff
@@ -96,6 +96,14 @@ await relayService.CreateNostrProfileAsync(nostrMetadata, nostrKeyHex, ok =>
             if (profileResult.IsFailure)
                 return Result.Failure<UpdateProjectProfileResponse>(profileResult.Error);
 
+            // Invalidate the local project cache so the founder's own app immediately
+            // reflects the new profile (name/banner/picture) on the next read instead
+            // of waiting for the stale-while-revalidate TTL.
+            var invalidateResult = await projectService.InvalidateCacheAsync(request.ProjectId);
```
dangershony: The invalidation of cache will only work for the founder not other users, this is not the best approach

PR #930, opened 2026-07-15
feat: recheck on-chain required amount against actual UTXO count

dangershony:
## Problem

Follow-up to #923 and companion to #929 (this one targeted at the **next** release).

The estimate shown to the user assumes a fixed number of inputs, but the signing path (`AddInputsFromAddressAndSignTransaction`) spends **all** UTXOs on the funding address and pays ~68 vB per input. If the payment arrives as multiple UTXOs (two separate sends, exchange withdrawal splits, RBF leftovers), the paid amount can no longer cover the miner fee — `MonitorAddressForFunds` declares the payment received, then `BuildInvestmentDraft` fails with "Insufficient funds".

#929 mitigates this with a static 3-input buffer; this PR makes the flow correct for any UTXO count.

## Fix

- `PaymentFlowConfig.OnChainRequiredForUtxoCount` — optional delegate that recomputes the required total for the actual number of detected UTXOs.
- `EstimateOnChainRequired` overload accepting an `inputCount` (existing signature delegates with count 1, so #929's padding merges cleanly on top).
- `PaymentFlowViewModel.PayToOnChainAddressAsync` — after funds arrive, recheck against the actual UTXO count; if short, show the exact shortfall ("Payment arrived in N parts — please send X sats more") and keep monitoring for the top-up instead of proceeding to a signing failure.
- `InvestPageViewModel` wires the delegate (invest flow only; deploy is unaffected as it doesn't restrict signing to the funding address).

Lightning/Boltz claims always deliver a single UTXO, so the recheck loop is a no-op on that path.

## Known caveat

The top-up monitor reuses `MonitorAddressForFunds`, which (via the mempool polling service) tracks unconfirmed UTXOs. If the first payment confirms before the top-up arrives, the aggregate check may need revisiting — worth verifying with a signet UAT run before merging. Marked as draft for the next release accordingly.

### DIFF
--- a/src/design/App/UI/Sections/FindProjects/InvestPageViewModel.cs
+++ b/src/design/App/UI/Sections/FindProjects/InvestPageViewModel.cs
@@ -706,6 +706,8 @@ public partial class InvestPageViewModel : ReactiveObject, IDisposable
             StageCount = Stages.Count,
             FeeRateSatsPerVbyte = 20,
             OnChainRequiredSatsOverride = PaymentFlowConfig.EstimateOnChainRequired(amountSats, Stages.Count, 20),
+            OnChainRequiredForUtxoCount = utxoCount =>
+                PaymentFlowConfig.EstimateOnChainRequired(amountSats, Stages.Count, 20, utxoCount),
             Title = $"Pay to {actionVerb}",
             SuccessTitle = successTitle,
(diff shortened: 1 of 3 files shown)


dangershony:
I am not convinced this commit should go in

PR #939, opened 2026-07-19
UI fixes + headless layout-regression test suite for all section views

Zazawowow:
## Summary

- **Headless UI layout-regression suite** (`src/design/App.Test.Integration/LayoutRegression/`): renders section views with real theme + Skia text metrics at breakpoint widths (360/390/767/768/1023/1280) and asserts no overlapping siblings, no children overflowing parent panels, and responsive subscriptions surviving detach/re-attach. 92 tests, ~11s, fully offline. Wired into CI.
- **Coverage**: Home, FindProjects, ProjectDetail, InvestPage, Portfolio, Funds, Settings, CreateProject (all 6 wizard steps), EditProfile, Funders, ManageProject, InvestmentDetail, ProjectCard — all with worst-case data (long names, many stages, all badges visible).
- **UI bug fixes found by the new tests**:
  - Settings: currency ComboBox row, Export/Show Logs buttons and Danger Zone subtitle overflowed their cards on phone widths — now wrap
  - FindProjects: search error text never wrapped (horizontal StackPanel gave it infinite width) — converted to Grid
  - InvestPage: long quick-amount labels could push buttons past their UniformGrid cell — MinWidth=0 + CharacterEllipsis
- **Other UI fixes** (first commit): founder loading issues, Funders/ManageProject/ProjectDetail responsive fixes
- Chore: removed accidentally committed .DS_Store files and session notes; gitignored

## Test plan
- `dotnet test src/design/App.Test.Integration/App.Test.Integration.csproj --filter "FullyQualifiedName~LayoutRegression"` — 92/92 passing (verified after rebase onto main)
- Before the fixes, the new tests failed with 9 real overlap/overflow violations; all assertions that failed now pass
- Desktop app builds and runs

PR #942, opened 2026-07-20
Founder scan resilience + derivation fixes (split from #939)

dangershony:
Splits the SDK and shared-library changes out of #939 so the UI/test PR stays focused.

### ScanFounderProjects.cs
- Self-heal: detects collapsed derived project keys (ARM64 JIT derivation bug) and rebuilds them from the seed
- Adds diagnostics logging for derivation collapse

### DocumentProjectService.cs
- Indexer lookup timeout extended 10s -> 30s (mobile networks regularly exceeded 10s, making founder scans come back empty on phones)
- Logs indexer lookup failures instead of silently dropping them
- Relay outages now degrade to returning local results instead of failing the whole call

### Angor.Shared
- New Bip32PublicDerivationHelper
- DerivationOperations fixes

Split out of #939 per review feedback.


dangershony:
this PR will probably not get pushed

PR #949, opened 2026-07-23
UI polish, Android relaunch crash fix, and parallel investment scanning

Zazawowow:
### Android
- Adaptive launcher icon: mipmap-based icon with monochrome foreground and Angor green (#5FAF78) background, plus round icon variant
- Fix crash on activity relaunch (config change outside ConfigurationChanges): cached section views were still parented to the dead SectionPanel, causing "already has a visual parent" — detach before re-adding

### UI
- New `i|Icon.Spin` style (infinite 360° rotation) applied to the syncing arrows-rotate icons in FundersView and ProjectCard
- MyProjects: on mobile the sidebar is now hidden entirely instead of rendered chrome-less (test updated to match new intended behavior)
- FundersView: bottom padding on expanded investor row
- Fix black icons on first launch in dark mode: Optris i:Icon bakes its Foreground at construction before theme dictionaries are reachable; force one theme-variant flip once startup theming settles

### SDK perf — faster "Syncing funding data…"
- `ProjectInvestmentsService`: parallelize spent-fund checks (Task.WhenAll instead of O(funders × stages) sequential awaits) and fetch tx hex + info concurrently
- `ProjectStatistics`: run investment scan and project fetch concurrently

## Testing
- `dotnet build src/design/App.Desktop` — clean
- Layout-regression suite: 92/92 passed
- SDK tests: 333 passed, 14 skipped


src/sdk/Angor.Sdk/Funding/Projects/ProjectInvestmentsService.cs
```diff
@@ -107,6 +107,13 @@ private async Task<Result<IEnumerable<StageData>>> ScanDynamicTypeInvestments(Pr
         // Dictionary to group stages by their release date
         var stagesByDate = new Dictionary<DateTime, StageData>();
 
+        // Collect all stage work items first so the spent-fund checks (each an
+        // indexer round-trip) can run in parallel instead of O(funders × stages)
+        // sequential awaits — this dominated the "Syncing funding data…" time.
+        var workItems = new List<(QueryTransactionOutput qouts, Transaction trx, int stageIndex,
```
dangershony: this code we need to check carefully that it didnt break anything

dangershony: Split this change out to #951 per the concern about unbounded parallelism at scale (hundreds of investors x stages). PR #949 now reverts ProjectInvestmentsService.cs to main; #951 re-applies the parallelization with a SemaphoreSlim bound (12 concurrent indexer requests) and a unit test asserting the max in-flight cap.

PR #951, opened 2026-07-24
perf: parallelize investment scanning with bounded concurrency

dangershony:
﻿Split out of #949 (per review discussion on the `ProjectInvestmentsService` change) so the scan-perf work gets focused review.

## Problem
`ScanDynamicTypeInvestments` awaited each spent-fund check sequentially — O(funders × stages) indexer round-trips dominated the "Syncing funding data…" time. But naively parallelizing with unbounded `Task.WhenAll` (as originally proposed in #949) is risky: projects can have **hundreds of investors**, and Fund projects multiply that by stage count — thousands of simultaneous HTTP requests would risk socket exhaustion and indexer rate limiting.

## Changes
- `ScanDynamicTypeInvestments`: collect all stage work items first, then run spent-fund checks in parallel with **bounded concurrency** (`RunBoundedAsync`, SemaphoreSlim capped at 12 concurrent indexer round-trips). Ordering is preserved (work-item index), so date-bucket grouping is byte-for-byte identical to the sequential version.
- `GetProjectInvestmentsTransactionsAsync`: fetch tx hex + tx info concurrently per investment; the per-investment fan-out (previously **already unbounded** `Task.WhenAll` over all investors) now also goes through the bound.
- `ScanInvestTypeInvestments`: the pre-existing unbounded per-stage `Task.WhenAll` is bounded too.

Bounded parallelism captures nearly all the latency win (it comes from overlapping round-trips, not N-way parallelism) with none of the failure risk at scale.

## Testing
- New unit test `ScanFullInvestments_ManyFundInvestors_LimitsConcurrentIndexerRequests`: 30 Fund investors × 3 stages with instrumented mocks tracking max in-flight indexer calls — asserts overlap happens (>1) and never exceeds 2 × cap (hex+info pairs share a slot).
- Existing grouping regression tests (`FundInvestorsWithShiftedSchedules`, `FundMixedBucket`) unchanged and green — confirms functional equivalence of the restructured loop.
- Full SDK suite: 334 passed, 14 skipped.

## Suggested UAT validation
`MultiFundClaimAndRecoverTest` (dynamic scan end-to-end incl. spent-type discrimination) and `MultiInvestClaimAndRecoverTest` (Invest path + FundersView).

### DIFF
--- a/src/sdk/Angor.Sdk/Funding/Projects/ProjectInvestmentsService.cs
+++ b/src/sdk/Angor.Sdk/Funding/Projects/ProjectInvestmentsService.cs
@@ -16,6 +16,37 @@ public class ProjectInvestmentsService(IProjectService projectService, INetworkC
     IAngorIndexerService angorIndexerService, IInvestorTransactionActions investorTransactionActions,
     ITransactionService transactionService, ILogger<ProjectInvestmentsService> logger) : IProjectInvestmentsService
 {
+    // Cap on concurrent indexer round-trips. Projects can have hundreds of investors
+    // (and dynamic projects multiply that by stage count), so unbounded Task.WhenAll
+    // could fire thousands of simultaneous HTTP requests — enough to exhaust sockets
+    // or trip indexer rate limits. A small degree of parallelism captures nearly all
+    // of the latency win from overlapping round-trips without that risk.
+    private const int MaxConcurrentIndexerRequests = 12;
+
+    /// <summary>
+    /// Runs the given task factories with bounded concurrency, preserving input order in the result.
+    /// </summary>
+    private static async Task<TResult[]> RunBoundedAsync<TResult>(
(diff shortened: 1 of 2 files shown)


src/sdk/Angor.Sdk/Funding/Projects/ProjectInvestmentsService.cs
```diff
@@ -66,13 +97,14 @@ private async Task<Result<IEnumerable<StageData>>> ScanInvestTypeInvestments(Pro
 
         foreach (var stage in stageDataList)
         {
-            var tasks = investmentsResult.Value.Select(tuple =>
+            var taskFactories = investmentsResult.Value.Select(tuple =>
```
dangershony, line 100: the stage.StageIndex + 2 is suspicious here (needs more investigation)

PR #958, opened 2026-09-12
fix: dust change output and sat/vB fee units in project deploy

dangershony:
﻿## What happened

A user reported project deployment failing with the generic banner **"The deployment transaction was prepared, but the network rejected the broadcast."** on mainnet.

Pulled their log export off Blossom (`src/tools/decrypt-log/fetch-latest.mjs`). The actual error was never shown in the UI:

```
[ERR] MempoolSpaceIndexerApi: Broadcast attempt 1/3 failed: Bad Request
  {"error":"sendrawtransaction RPC error: {\"code\":-26,\"message\":\"dust\"}"}
```

Sequence from the log:

| Time | Event |
|---|---|
| 16:45:31 | `Not enough funds, expected 0.00010001 BTC, found 0.0001 BTC` — funded exactly 10,000 sats, `AngorCreateFeeSats` is 10,001 |
| 16:50:58 | monitoring `bc1q89k3z...` for 10,000 sats |
| 16:52:01 | `Required amount met. Required: 10000, Available: 10348` — user topped up |
| 16:52:02 | broadcast rejected: **dust** (x3 retries, then again at 18:04) |

10,348 in − 10,001 (Angor fee output) − ~150 (miner fee) ≈ **~200 sat change output**, below the dust limit.

## Root cause

Three fixes already in the repo had never been applied to the deploy path.

**1. Dust change.** `12b18815` fixed `AddInputsFromAddressAndSignTransaction` to fold sub-dust change into the fee. Deploy goes through the *other* method, `AddInputsAndSignTransaction`, which hands off to the NBitcoin builder with `DustPrevention = false` and emits the dust output verbatim.

**2. Fee units.** `d39eb878` fixed the claim flow: the UI picker returns sat/vByte, NBitcoin's `FeeRate(Money)` is sat/kB. Deploy still passed the raw value, so "20 sat/vB" became 20 sat/kB and got clamped to the 1000 sat/kB floor — every deploy went out at ~1 sat/vB and the fee selector was a no-op.

**3. Coin selection.** `d67d72df` fixed `SendAmount` to select UTXOs covering amount + fee. Deploy still selected against the output total alone — that is the 16:45 `Not enough funds`.

## Changes

- **`WalletOperations.AddInputsAndSignTransaction`** (builder branch) — folds change `<= ProtocolConstants.DustThresholdSats` into the fee; re-selects UTXOs for `outputs + fee + dust` when the initial selection falls short; rebuilds with a *fresh* `TransactionBuilder` (`SendFees` and `SendEstimatedFees` accumulate on one instance — latent bug in the old code); verifies the built tx for dust change rather than trusting the projection, since the builder can pick a different coin subset. `DustPrevention` stays off, now with a comment explaining why (the zero-value OP_RETURN would be dropped).
- **`CreateProjectRequest.SelectedFeeRate`** — `long` to `DomainFeerate`, matching `BuildRecoveryTransaction` / `BuildPenaltyReleaseTransaction` / etc. Handler passes `.SatsPerKilobyte`. Propagated through `IProjectAppService`, `ProjectAppService`, and the three `DeployFlowViewModel` call sites. Carrying the unit in the type is what stops this recurring.
- **`DeployFlowViewModel`** — `DescribePrepareFailure` / `DescribeBroadcastFailure` map dust, low-fee, and already-spent-input rejections to actionable text across all 6 sites, falling back to the old message. Same approach as `SendFundsModal` in `d67d72df`.

## Tests

- `AddInputsAndSignTransaction_WhenChangeWouldBeDust_AddsItToTheFeeInsteadOfCreatingAnOutput` — reproduces the reporter's exact amounts. **Verified it fails against the pre-fix code** (stashed `WalletOperations.cs`, ran it, got `[FAIL]`, restored).
- `Handle_WhenSuccessful_PassesFeeRateToWalletOperationsInSatsPerKilobyte` — pins the 20 to 20,000 conversion.

| Suite | Result |
|---|---|
| `Angor.Shared.Tests` | 157 passed |
| `Angor.Sdk.Tests` | 336 passed, 14 skipped |
| `App.Test.Integration` (LayoutRegression) | 187 passed |
| `App.Desktop` + `WebApp.sln` builds | 0 errors |

`src/App.sln` as a whole needs the `ios` workload, unavailable on this machine. Pre-existing and unrelated.

## Not addressed here

- **The deploy flow requests exactly 10,000 sats** while the tx needs `AngorCreateFeeSats` + miner fee + dust-safe change. That is what caused the first failure at 16:45. Should be `10001 + estimatedFee + DustThresholdSats`.
- **`FeeSelectionPopup` presets are hardcoded** (50/20/5 sat/vB) and do not call `IWalletAppService.GetFeeEstimates()`, which already wraps mempool's `fastestFee`/`halfHourFee`/`economyFee`. The webapp uses live estimates in ~10 places; the new app only calls `GetFeeEstimates()` from `WalletAppService` and never from the deploy picker. Now that the unit bug is fixed the presets are applied literally, so 50 sat/vB is a real overpay on a ~150 vB tx.
- **`MempoolSpaceIndexerApi.cs:170`** builds its error as `ReasonPhrase + content` with no status code and no separator, producing strings like `Bad Request{"error":...}`.

### DIFF
--- a/src/design/App/UI/Sections/MyProjects/Deploy/DeployFlowViewModel.cs
+++ b/src/design/App/UI/Sections/MyProjects/Deploy/DeployFlowViewModel.cs
@@ -6,6 +6,7 @@ using Angor.Sdk.Funding.Founder;
 using Angor.Sdk.Funding.Founder.Operations;
 using Angor.Sdk.Funding.Projects;
 using Angor.Sdk.Funding.Projects.Dtos;
+using Angor.Sdk.Funding.Shared;
 using Angor.Sdk.Wallet.Application;
 using Angor.Sdk.Funding.Investor;
 using Angor.Sdk.Wallet.Domain;
@@ -243,11 +244,11 @@ public partial class DeployFlowViewModel : ReactiveObject
 
         // Step 4: Create blockchain transaction
         var txResult = await _projectAppService.CreateProject(
-            walletId, SelectedFeeRate, ProjectData, infoResult.Value.EventId, projectSeed);
+            walletId, new DomainFeerate(SelectedFeeRate), ProjectData, infoResult.Value.EventId, projectSeed);
         if (txResult.IsFailure)
         {
             _logger.LogError("Deploy: create blockchain transaction failed: {Error}", txResult.Error);
-            return Result.Failure("We could not prepare the on-chain deployment transaction. Check your wallet balance, fee rate, and spendable funds, then try again.");
+            return Result.Failure(DescribePrepareFailure(txResult.Error));
         }
 
         // Step 5: Publish to blockchain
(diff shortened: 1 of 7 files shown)


dangershony:
Closing — reworking with a minimal patch after review feedback.

PR #960, opened 2026-09-12
feat: fetch fee rates from the indexer instead of hardcoding them

dangershony:
Stacked on #959 — branched from it, so the diff here is only the fee-rate work. Merge #959 first.

## The problem

The `src/design` app never called the fee estimation API. Every fee value traced back to a literal:

| Location | Value |
|---|---|
| `FeeSelectionPopup.axaml.cs:140-147` | `priority => 50L, economy => 5L, _ => 20L` |
| `FeeSelectionPopup.axaml:158,175,193` | `Text="50 sat/vB"`, `"20 sat/vB"`, `"5 sat/vB"` |
| `DeployFlowViewModel.cs:60` | `selectedFeeRate = 20` |
| `PaymentFlowConfig.cs:20` | `FeeRateSatsPerVbyte = 20` |

Live mainnet right now, from the endpoint the code already knows how to call:

```
$ GET https://mempool.space/api/v1/fees/recommended
fastestFee: 1   halfHourFee: 1   hourFee: 1   economyFee: 1   minimumFee: 1
```

So "Standard" was 20x the going rate and "Priority" 50x.

This was invisible before #959, because the sat/vB-as-sat/kB bug meant all three presets fell under the 1000 sat/kB floor at `WalletOperations.cs:128` and collapsed to an identical 252 sats (1 sat/vB). #959 makes the selection real, which makes the wrong presets start to bite — the two should not ship apart.

## The plumbing already existed

`MempoolSpaceIndexerApi.GetFeeEstimationAsync` (`:334`) → `WalletOperations.GetFeeEstimationAsync` (`:738`) → `IWalletAppService.GetFeeEstimates` (`WalletAppService.cs:524`).

16 call sites across the repo, none in `src/design`:

| Project | Callers |
|---|---|
| `webapp/Angor.Client` | 10 |
| `avalonia/AngorApp` | 1 — `UIServices.cs:154` |
| `cli/Angor.Cli` | 2 |
| **`design/App`** | **0** |

`UIServices.GetFeeratePresetsAsync` fetches live rates and only falls back to literals in its `catch`. The rewrite kept the fallback and dropped the fetch, so this is a regression against both the webapp and the old Avalonia app.

## Changes

**New `IFeeRateProvider`** (`UI/Shared/Services/FeeRateProvider.cs`)
- Fetches via `IWalletAppService.GetFeeEstimates()`
- Maps sat/kB estimates keyed by confirmations (1/3/6/18) onto Priority/Standard/Economy in sat/vB, clamped to ≥1 — same mapping as `UIServices.cs:150-170`
- 60s cache, 3s timeout, `SemaphoreSlim` so concurrent callers make one request
- Never throws; returns `FeeRates.Fallback` (the old 50/20/5) on failure
- Logs the resolved rates at Information, which also gives us the debugging hook that was missing when diagnosing the original report

**`FeeSelectionPopup.ShowAsync`** awaits the rates and applies them to the preset values *and* the `"N sat/vB"` labels via `ApplyRates`, so the number displayed and the number used cannot drift. This also gives the invest flow live presets, since `PaymentFlowView.axaml.cs:149` uses the same popup.

**`DeployFlowViewModel.Show`** sets `SelectedFeeRate` from `IFeeRateProvider.Current`.

This one matters more than it looks. The invoice path never opens the fee popup — `PayInvoiceInsteadButton` goes straight to `ShowPayFee()` (`DeployFlowOverlay.axaml.cs:77-79`) — so `Show` is the only place its rate is chosen, and #959 budgets the invoice amount from it. That is the exact path the original bug report came through (`DeployAfterPaymentAsync` → `RunDeployStepsAsync` in their log). Wiring only the popup would not have touched it.

`Show` stays **synchronous** deliberately: the view click handler calls it via `Deploy()`, and ten integration tests call `wizardVm.Deploy()` then immediately assert on `PaymentFlow` (`CreateProjectTest.cs:336` — *"PaymentFlow should be created by DeployFlow.Show()"*). Making it async would have rippled through all of them. Instead the provider is warmed in the `DeployFlowViewModel` constructor — which runs when the wizard opens, many steps before deploy is reachable — and `Show` reads `Current`.

## Tests

`FeeRateProviderMapTests` covers the pure mapping:
- sat/kB → sat/vB conversion and correct target selection (40/20/5)
- clamp to 1 sat/vB, using the real quiet-mainnet case (`1 * 1100 = 1100` sat/kB)
- fallback on empty and on all-zero responses
- collapsing onto the closest target when the indexer omits confirmations

| Suite | Result |
|---|---|
| `App.Test.Integration` (Layout + FeeRate + panels) | 203 passed |
| `Angor.Sdk.Tests` | 335 passed, 14 skipped |
| `Angor.Shared.Tests` | 156 passed |
| `App` Release build | 0 errors |

One caveat worth recording: on the first combined run `FindProjectsPanelTests.FindProjectsFlow_WithWallet_WalletSelectorAndInvoice` failed, then passed on an isolated run and again on an identical combined re-run (203/203). That test asserts *"projects should load from SDK within timeout - ensure testnet indexer/relays are reachable"* and touches no code in this PR, so it reads as indexer/relay flakiness rather than a regression — but I did not capture the assertion text on that first run, so I can't prove it. Flagging rather than burying it.

## Scope notes

- `InvestPageViewModel.cs:706,744` still hardcodes `20` for its `EstimateOnChainRequired` budget and `DomainFeerate`. The invest flow gets live values in the *popup*, but not in its invoice budget. Same fix applies; left out to keep this reviewable.
- `PaymentFlowConfig.FeeRateSatsPerVbyte = 20` remains as a default, now overridden by the deploy flow.
- `App.Services.GetRequiredService<IFeeRateProvider>()` in `ShowAsync` follows the existing service-locator pattern in this codebase (`SendFundsModal.axaml.cs:43`, `WalletDetailModal.axaml.cs:42`) since the popup is constructed by hand rather than through DI.


dangershony:
Rebased onto `main` now that #959, #961 and #962 are merged. All three were squash-merged, so I replayed only this branch''s own commit with `git rebase --onto origin/main c471ea17` — clean, no conflicts. Base auto-retargeted to `main`, diff is now purely the fee-rate work (5 files, +311/-5).

Re-verified on top of merged `main`:

| Suite | Result |
|---|---|
| `Angor.Sdk.Tests` | 346 passed, 14 skipped |
| `Angor.Shared.Tests` | 156 passed |
| `App.Test.Integration` (Layout + FeeRate + PlatformBack) | 199 passed |
| UAT `MultiFundClaimAndRecoverTest` | Passed |
| UAT `CreateProjectTest` + `MultiInvestClaimAndRecoverTest` | Passed |

`MultiFundClaimAndRecoverTest` is the one that failed before #961 — it passes now with #961''s witness-placeholder fixes in `main`, confirming the dependency is satisfied.

### Note on how UAT was run

`signet.angor.online` — the **primary** signet indexer — is currently unreachable (DNS resolves to 15.235.3.238, but TCP 443 times out after 15s). UAT initially failed at wallet funding with *"Wallet balance did not become non-zero in time"*, with no relay or dust errors anywhere in the child logs.

The runs above used `ANGOR_INDEXER_URL=https://test.indexer.angor.io`, which is the only signet indexer currently responding. That is an infrastructure outage, unrelated to this PR, and is being handled separately.

